PR #4, opened 2026-03-25
[ci] Add Ubuntu arm jobs to ci

mcbarton:
cppyy doesn't have as many tests passing on Ubuntu arm as Ubuntu x86 I believe, so this PR adds Ubuntu arm jobs to the ci.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,18 @@ jobs:
             llvm: 21
             python: "3.14"
             runtime_cxx_standard: 20
+          - os: ubuntu-24.04-arm
+            llvm: 20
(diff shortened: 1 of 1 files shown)


aaronj0:
> cppyy doesn't have as many tests passing on Ubuntu arm as Ubuntu x86 I believe, so this PR adds Ubuntu arm jobs to the ci.

Hi, I have a slightly different CI setup in mind that I'm working on. Currently, we are still continuing active development on the forks, so the goal is not to extensively cover every platform, but to consolidate what the package will look like and work out some engineering details with the ROOT migration. Would be great if you could pick this up once we have our first pre-release in, and the forks are archived

PR #6, opened 2026-06-10
Migrate to compiler-research/ci-workflows

aaronj0:
Plus some minor developments, extending to a Cling backend and adding configuration files for a modular CI setup. This system moves to using the reusable actions from ci-workflows () and creates a build matrix for PR runs and extended coverage for nightlies.  This PR will be followed up with the pre-commit and lint workflows before opening a PyPI workflow to build Python wheels. Note that the std::span test fails for me locally, also with the 3-repo setup so it needs to be resolved before deprecating the forks.

Depends on http://github.com/compiler-research/ci-workflows/pull/80


.github/workflows/ci.yml
```diff
@@ -1,96 +1,57 @@
...
-        env:
-          CPPINTEROP_EXTRA_INTERPRETER_ARGS: "-std=c++${{ matrix.runtime_cxx_standard }}"
-
-      - name: Upload result
-        if: always()
-        uses: actions/upload-artifact@v4
-        with:
-          name: result-${{ matrix.os }}-llvm${{ matrix.llvm }}-py${{ matrix.python }}-cpp${{ matrix.runtime_cxx_standard }}
-          path: test-output.txt
+          - { os: ubuntu-24.04,   llvm: '21', flavor: system, py: '3.14', cxx: '20', vg: true  }
```
vgvassilev: We need `name:` here because it duplicates the strings and in the github ci bar it shows twice.

.github/workflows/ci.yml
```diff
@@ -1,96 +1,57 @@
...
+          - { os: ubuntu-24.04,   llvm: '22', flavor: '',     py: '3.14', cxx: '20', vg: false }
+          - { os: macos-26,       llvm: '21', flavor: system, py: '3.14', cxx: '20', vg: false }
+          - { os: macos-26-intel, llvm: '21', flavor: system, py: '3.14', cxx: '20', vg: false }
+          # cling backend, against the cached llvm-root (cling-llvm20) recipe cell.
+          - { os: ubuntu-24.04,   llvm: '20', flavor: cling, fv: cling-llvm20, cling: true, py: '3.14', cxx: '20', vg: false }
```
vgvassilev: What is `fv`, isn't `cling:true` redundant?

.github/workflows/ci.yml
vgvassilev: `vg:false` should be default and omitted.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,92 @@
...
+    # NOTE: retarget to compiler-research/ci-workflows/...@main once upstreamed.
+    with:
+      os: ${{ matrix.os }}
+      llvm-version: ${{ matrix.llvm }}
+      llvm-flavor: ${{ matrix.flavor }}
+      llvm-flavor-version: ${{ matrix.fv || '' }}
+      python-version: ${{ matrix.py }}
+      cxx-standard: ${{ matrix.cxx }}
+      valgrind: ${{ matrix.vg }}
+      run-sweeps: true
```
vgvassilev: This name makes no sense -- maybe run-all-crashing-tests or something is better.

cmake/AddCppInterOp.cmake
```diff
@@ -0,0 +1,76 @@
+# Configures the CppInterOp ExternalProject. Default backend is clang-repl;
```
vgvassilev, line 1: Why do we need this file? If it has something to do with the ci, then then it defeats the purpose of the ci-workflows changes.

.github/workflows/ci.yml
aaronj0: Addressed

aaronj0: Done

cmake/AddCppInterOp.cmake
aaronj0: This was discussed and is resolved (AddCppInterOp.cmake)

.github/workflows/nightly.yml
aaronj0: Addressed in https://github.com/compiler-research/ci-workflows/pull/83

.github/workflows/ci.yml
aaronj0: This passes what the ci-workflows `setup-llvm` expects as a "recipe" (`cling-llvm20`)

aaronj0:
Parts of the review comments are addressed in https://github.com/compiler-research/ci-workflows/pull/83

vgvassilev (approved):
Lgtm!

PR #7, opened 2026-06-29
Update to work with latest CppInterOp and strong types

aaronj0:
mirrors #6 but with a syncup, that should update to the latest HEAD, fixing the visible CI failures in the process (std::span converter,  LLVM22). I will then rebase that PR to only include the CI changes


aaronj0:
merging, as the content in this PR was approved on #6, the rest is a routine sync with no changes on top.

PR #8, opened 2026-07-07
Syncup source with cppyy forks

aaronj0:
This pulls in the latest commits fixing the valgrind issue in the cppyy tests (use  ll.array_delete instead of ll.free).
Once this lands, I can rebase #6 on top, to then land the new CI system.


aaronj0:
Merging as this is a routine syncup and does not diverge from the forks.

PR #11, opened 2026-07-19
[ci] Use source LLVM on ARM valgrind nightlies

aaronj0:
Mirrors what is currently done in CppInterOp and the cppyy repos for ARM (use source built LLVM 22 instead of apt llvm 21for valgrind). Fixes the nightly failures


vgvassilev (approved):
Lgtm!

PR #16, opened 2026-08-14
[cppinterop] Bump latest supported version for Cling-LLVM22

aaronj0:


aaronj0:
Cling-LLVM22 is now supported cc @vgvassilev @guitargeek

PR #17, opened 2026-08-14
[ci][test] Improve requirements.txt, enable psutil and numba

aaronj0:


aaronj0:
Numba has dropped OS X x86 support since 0.61. Hence, this PR locks the requirements.txt to enforce the supported version window.

PR #18, opened 2026-08-14
Support developer builds of CppInterOp/LLVM, add project verification and robust fallbacks

aaronj0:
Developing CppInterOp or LLVM against cppjit requires building from local trees instead of the pinned clone. `CPPINTEROP_SOURCE_DIR` now builds CppInterOp from a local checkout with incremental rebuilds, and a post-config check rejects source trees that are not CppInterOp. Validated against a fetch of ci-workflows LLVM 22 and an uninstalled LLVM 21.1.8 build tree. This change also improves how robust the build system is with various permutations of developer setups (custom CppInterOp/LLVM development)

The behavior of cppjit's build system with this change is:

| `pip install . --config-settings=cmake.define.<...>` | CppInterOp | LLVM |
|---|---|---|
| none (basic install) | clones and builds the pinned `CPPINTEROP_GIT_TAG` | `find_package(LLVM)`: `$CONDA_PREFIX` hint, then the system search |
| `LLVM_DIR=<dir>` | clones and builds the pin | uses provided install prefix or build tree only, no fallback as this is a developer override |
| `CPPINTEROP_GIT_REPOSITORY=<url>` / `CPPINTEROP_GIT_TAG=<ref>` | clones the given remote/ref and builds it | per the rows above |
| `CPPINTEROP_SOURCE_DIR=<dir>` | builds the local tree in place with incremental rebuilds (output is streamed) | per the rows above |

Priority:
- `CPPINTEROP_SOURCE_DIR` over `CPPINTEROP_GIT_REPOSITORY`/`CPPINTEROP_GIT_TAG`
- `LLVM_DIR` over the `$CONDA_PREFIX` hint over the system search.

The two axes combine freely, and the selected LLVM is forwarded to the CppInterOp sub-build.

The negative cases each fail at config time with actionable feedback:

| Case | Result |
|---|---|
| no LLVM found | `No LLVM CMake package found. Install LLVM 20-22 development packages (apt: llvm-22-dev libclang-22-dev; conda: llvmdev clangdev), or point cppjit at your own LLVM build with -DLLVM_DIR=...` |
| `LLVM_DIR` invalid | `No LLVMConfig.cmake under LLVM_DIR (<path>); expected <install prefix or build tree>/lib/cmake/llvm` |
| `Clang_DIR` invalid | `No ClangConfig.cmake under Clang_DIR (<path>)` |
| LLVM outside 20-22 | `LLVM <version> is unsupported: the currently supported CppInterOp version (CPPINTEROP_GIT_TAG) only supports LLVM 20-22` |
| `CPPINTEROP_SOURCE_DIR` is not a CMake tree | the sub-configure reports the missing CMakeLists.txt |
| the source or ref is not CppInterOp | pre-build, post scikit-build-core config: `The provided source path override declares project '<name>', not CppInterOp`|

The supported range is defined once next to the pin (`CPPJIT_LLVM_VERSION_MIN`/`MAX`, updated together with the supported CppInterOp version), and the Clang lookup is now confined to the discovered LLVM's own package dir, preventing a mismatched system clang from breaking the config on a machine with several toolchains installed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,8 @@ include(GNUInstallDirs)
 option(CPPJIT_ENABLE_CPPINTEROP_TESTS "enable CppInterOp tests" OFF)
 set(CPPINTEROP_GIT_REPOSITORY "https://github.com/compiler-research/CppInterOp.git" CACHE STRING "")
 set(CPPINTEROP_GIT_TAG "8d624c621a4b95e36ff73ac708c85a768287478f" CACHE STRING "")
+set(CPPINTEROP_SOURCE_DIR "" CACHE PATH
+    "Override default CppInterOp built by ExternalProject_Add, with a path to local CppInterOp source")
 
 set(Python_FIND_VIRTUALENV ONLY)
 find_package(Python COMPONENTS Interpreter Development)
@@ -64,10 +66,8 @@ if(Clang_FOUND)
     message(STATUS "Found Clang at ${Clang_DIR}")
 endif()
 
-# build CppInterOp, we should expose the option to use pre-installed CppInterOp for example using conda
-
-# CPPINTEROP_DIR should resolve to the final installed location at runtime.
-# ask for site-packages path; fall back to CMAKE_INSTALL_PREFIX for standalone builds.
+# CppInterOp is installed at the location cppjit ships at runtime: ask for the
+# site-packages path; fall back to CMAKE_INSTALL_PREFIX for standalone builds.
 execute_process(
     COMMAND ${Python_EXECUTABLE} -c "import sysconfig; print(sysconfig.get_path('platlib'))"
     OUTPUT_VARIABLE _python_platlib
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.16)
+cmake_minimum_required(VERSION 3.21)
```
guitargeek: What do we need CMake 3.21 for? It would be nice to not require a version higher than 3.20 for now, to be consistent with LLVM:

https://github.com/llvm/llvm-project/blob/main/clang/CMakeLists.txt#L1

guitargeek (approved):
LGTM! Consider CMake 3.20 compatibility if possible, but if it can't be done so be it. LLVM will also require CMake 3.31, starting from LLVM 24.

CMakeLists.txt
aaronj0: Yes we can proceed with cmake 3.20, thanks!

PR #21, opened 2026-08-17
Load CppInterOp explicitly and resolve it relative to libcppjit

aaronj0:
The backend used to dlopen `libclangCppInterOp` inside a global constructor and locate it through absolute build-time paths, so a wheel built on one machine could not load on another. A constructor also runs under the loader lock, inside the dlopen of `libcppjit` itself: the nested dlopen re-enters the loader on partially consistent state, runs the inner library's initializers with no ordering guarantees, and holds the lock through the entire interpreter and JIT setup, deadlocking any other thread that touches the loader (as reported by @Vipul-Cariappa)

The ctor also had no error channel: on failure, the old code only prints and leave a half-initialized library. With this patch the explicit `LoadCppInterOp()` runs after loading completes and returns a status that turns into a proper exception on the Python side.

This PR implements the following:
- CppInterOp loading and interpreter setup now run in an exported, idempotent `LoadCppInterOp()` that `_cpython_cppjit.py` calls before the extension module import
- Library and headers are resolved relative to libcppjit's own load location, with the build-time install prefix as the fallback for build trees. This is necessary for the wheels building infrastructure (https://github.com/compiler-research/cppjit/tree/wheels-build-test-ci), which will be added as a separate PR.
- Dead fork-inherited code and the fake cling environment (unused include paths, cling::runtime::gCling) are removed along the way, NFC

Validation:
- full test suite passes locally across different cxx standard builds
- wheel relocatability: a wheel built in an isolated environment installs and runs in a second fresh venv after the build
environment is deleted
- the install-prefix fallback and repeated LoadCppInterOp() calls verified against a bare build-directory `libcppjit.so`
- clang-format-18 and ruff clean

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,10 +95,11 @@ execute_process(
     OUTPUT_STRIP_TRAILING_WHITESPACE
 )
 if(_python_platlib)
-    set(CPPINTEROP_INSTALL_DIR "${_python_platlib}/cppjit_backend")
+    set(CPPINTEROP_INSTALL_PREFIX "${_python_platlib}")
 else()
-    set(CPPINTEROP_INSTALL_DIR "${CMAKE_INSTALL_PREFIX}/cppjit_backend")
+    set(CPPINTEROP_INSTALL_PREFIX "${CMAKE_INSTALL_PREFIX}")
 endif()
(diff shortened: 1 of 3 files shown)


aaronj0:
As per @vgvassilev's request, I've broken this into atomic PRs that makes the changes easy to review.

PR #22, opened 2026-08-17
Support building wheels for manylinux and OSX

aaronj0:
Rebased on top of #21 since we need that patch for relocatable wheels.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -1,0 +1,114 @@
+name: Wheels
+
+# Build the wheels (cibuildwheel; config in pyproject.toml) and the sdist
+# as artifacts. setup-recipe stages the llvm-wheel toolchain at /opt/llvm;
+# linux mounts it into the build container, the same manylinux_2_28 image
+# the toolchain was built on.
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - '.github/workflows/wheels.yml'
+      - '.github/wheel_smoke.py'
+      - 'pyproject.toml'
+      - 'CMakeLists.txt'
+      - 'cmake/**'
+      - 'src/interop/**'
+      - 'python/cppjit/_cpython_cppjit.py'
+  push:
+    tags: ['v*']
+  schedule:
+    - cron: '30 4 * * 1'
+
+permissions:
+  contents: read
+
+concurrency:
+  group: wheels-${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  wheels:
+    name: wheels ${{ matrix.label }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - { os: ubuntu-24.04, label: manylinux-x86_64, arch: x86_64 }
+          - { os: macos-26,     label: macosx-arm64,     arch: arm64 }
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@v7
+
+      # ref pins the recipe content the cache key is computed from.
+      - uses: compiler-research/ci-workflows/actions/setup-recipe@main
+        id: llvm
+        with:
+          recipe: llvm-wheel
+          version: '21.1.8'
+          os: ${{ matrix.os }}
+          arch: ${{ matrix.arch }}
+          ref: b760e4c171961786b7b20e2cc514302df5373eef
+
+      - name: Stage the toolchain at /opt/llvm
+        env:
+          RECIPE_PATH: ${{ steps.llvm.outputs.path }}
+        run: sudo mv "$RECIPE_PATH" /opt/llvm
+
+      - uses: pypa/cibuildwheel@v4.2.0
+
+      - uses: actions/upload-artifact@v7
+        with:
+          name: wheels-${{ matrix.label }}
+          path: wheelhouse/*.whl
+
+  sdist:
+    name: sdist
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v7
+
+      - run: pipx run build --sdist
+
+      - uses: actions/upload-artifact@v7
+        with:
+          name: sdist
+          path: dist/*.tar.gz
+
+  # Run the full suite on a plain runner, outside the manylinux
+  # container the wheel was built in.
+  test-wheel:
+    name: test wheel (full suite)
+    needs: wheels
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: actions/setup-python@v7
+        with:
+          python-version: '3.12'
+
+      - uses: actions/download-artifact@v8
+        with:
+          name: wheels-manylinux-x86_64
+          path: wheelhouse
+
+      - name: Install the test suite's native deps
+        # test_eigen/test_boost need them; the CI cells install the same pair.
+        run: sudo apt-get -q update && sudo apt-get -y install libeigen3-dev libboost-dev
+
+      - name: Install the wheel and the test requirements
+        run: python -m pip install wheelhouse/cppjit-*cp312*.whl -r requirements.txt
+
+      - name: Smoke the wheel outside pytest
+        run: python -X faulthandler .github/wheel_smoke.py
+
+      - name: Run the test suite against the installed wheel
+        env:
+          CPPINTEROP_EXTRA_INTERPRETER_ARGS: -std=c++20
+        run: |
+          cd test
+          make -j$(nproc) PYTHON=python
+          python -m pytest -ra
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,7 +6,7 @@ build-backend = "scikit_build_core.build"
 name = "cppjit"
 dynamic = ["version"]
 description = "CppJIT: fast and automatic Python-C++ interoperability"
-license = {text = "LBNL BSD"}
+license = "BSD-3-Clause-LBNL"
 requires-python = ">=3.12"
 authors = [
     {name = "Aaron Jomy"},
@@ -21,6 +21,7 @@ maintainers = [
 ]
 
 [tool.scikit-build]
+minimum-version = "build-system.requires"
 wheel.install-dir = "."
 wheel.packages = ["python/cppjit"]
 cmake.build-type = "Release"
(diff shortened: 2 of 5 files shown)


aaronj0:
@vgvassilev A note: The cleanest way to do this is to have wheel-builder images on ci-workflows that provision the image and LLVM build for manylinux and OS X. I will open a PR once I have a working draft locally and update the approach here to trigger that common workflow. That can then, in principle, be reused to build wheels for any other Python package based on LLVM under compiler-research

pyproject.toml
```diff
@@ -30,6 +30,38 @@ provider = "scikit_build_core.metadata.regex"
...
+# LLVM/Clang 21 archives, pinned by sha256) at /opt/llvm.
+before-all = "bash {project}/.github/cibw_before_all_linux.sh"
+# CMAKE_ARGS reaches the CppInterOp ExternalProject sub-configure.
+environment = { CMAKE_ARGS = "-DLLVM_DIR=/opt/llvm/lib/cmake/llvm -DClang_DIR=/opt/llvm/lib/cmake/clang" }
+
+[tool.cibuildwheel.macos]
+archs = ["arm64"]
+# eigen and boost are test-suite build deps; the LLVM toolchain is the
+# llvm-wheel artifact wheels.yml stages at /opt/llvm before this runs.
+before-all = "brew install eigen boost"
```
mcbarton: Do this assume you are using `homebrew`? Not everyone does use it on MacOS. I use Macports for example.

pyproject.toml
```diff
@@ -30,6 +30,38 @@ provider = "scikit_build_core.metadata.regex"
...
+archs = ["x86_64"]
+manylinux-x86_64-image = "manylinux_2_28"
+# before-all stages the ci-workflows llvm-wheel toolchain (static
+# LLVM/Clang 21 archives, pinned by sha256) at /opt/llvm.
+before-all = "bash {project}/.github/cibw_before_all_linux.sh"
+# CMAKE_ARGS reaches the CppInterOp ExternalProject sub-configure.
+environment = { CMAKE_ARGS = "-DLLVM_DIR=/opt/llvm/lib/cmake/llvm -DClang_DIR=/opt/llvm/lib/cmake/clang" }
+
+[tool.cibuildwheel.macos]
+archs = ["arm64"]
```
mcbarton, line 49: I know its being supported less and less, but what about MacOS intel. There is clearly appetite for it by people wanting to use cppjit (see https://github.com/compiler-research/cppjit/issues/39#issue-5234755449)

pyproject.toml
```diff
@@ -30,6 +30,38 @@ provider = "scikit_build_core.metadata.regex"
...
+# eigen and boost are test-suite build deps; the LLVM toolchain is the
+# llvm-wheel artifact wheels.yml stages at /opt/llvm before this runs.
+before-all = "brew install eigen boost"
+# The macOS test phase runs on the runner itself, so the standard suite
+# runs here; the linux equivalent is the test-wheel job.
+test-command = "python -m pip install -r requirements.txt && python .github/wheel_smoke.py && cd test && make -j$(sysctl -n hw.ncpu) PYTHON=python && python -m pytest -ra"
+# 14.0 is the toolchain's own deployment floor (set at llvm-wheel build
+# time); nothing brew-built ships in the wheel.
+environment = { CMAKE_ARGS = "-DLLVM_DIR=/opt/llvm/lib/cmake/llvm -DClang_DIR=/opt/llvm/lib/cmake/clang", MACOSX_DEPLOYMENT_TARGET = "14.0" }
```
mcbarton, line 52: I'm a little confused. Where does this `/opt/llvm/` build come from?

pyproject.toml
```diff
@@ -30,6 +30,38 @@ provider = "scikit_build_core.metadata.regex"
...
+test-command = "python .github/wheel_smoke.py"
+
+[tool.cibuildwheel.linux]
+archs = ["x86_64"]
+manylinux-x86_64-image = "manylinux_2_28"
+# before-all stages the ci-workflows llvm-wheel toolchain (static
+# LLVM/Clang 21 archives, pinned by sha256) at /opt/llvm.
+before-all = "bash {project}/.github/cibw_before_all_linux.sh"
+# CMAKE_ARGS reaches the CppInterOp ExternalProject sub-configure.
+environment = { CMAKE_ARGS = "-DLLVM_DIR=/opt/llvm/lib/cmake/llvm -DClang_DIR=/opt/llvm/lib/cmake/clang" }
```
mcbarton, line 46: Same to MacOS quesion about the `/opt/llvm/` build

pyproject.toml
```diff
@@ -30,6 +30,38 @@ provider = "scikit_build_core.metadata.regex"
...
+build-verbosity = 1
+# Every wheel gets a smoke import; the standard suite runs where the
+# test env is a real runner — macOS below, linux in the test-wheel job
+# on a plain runner (the manylinux container's internal CPython aborts
+# mid-suite).
+test-sources = ["test", "requirements.txt", ".github/wheel_smoke.py"]
+test-command = "python .github/wheel_smoke.py"
+
+[tool.cibuildwheel.linux]
+archs = ["x86_64"]
```
mcbarton, line 42: What about Linux arm?

.github/cibw_before_all_linux.sh
```diff
@@ -1,13 +1,19 @@
...
 set -euo pipefail
 
-curl -fsSL --retry 5 -o /tmp/micromamba.tar.bz2 https://micro.mamba.pm/api/micromamba/linux-64/latest
-tar -xjf /tmp/micromamba.tar.bz2 -C /usr/local bin/micromamba
-export MAMBA_ROOT_PREFIX=/opt/mamba
-# zlib must be explicit: llvmdev depends only on libzlib (the bare .so),
-# and linking LLVMSupport fails at generate time unless find_package(ZLIB)
-# resolves, which needs zlib.h from the zlib package.
-micromamba create -y -p /opt/llvm -c conda-forge 'llvmdev=21.*' 'clangdev=21.*' lld zlib
+asset="llvm-wheel-21.1.8-ubuntu-24.04-x86_64-12af669f7a4444de.tar.zst"
```
vgvassilev: I am still not sure if I understand the special llvm-wheel. How is it different from what we vendor?

aaronj0:
@mcbarton A comment: this PR is in development and some of the decisions here are still in discussion. I can request your review at a point things have converged. To answer your current questions:

Yes, Intel mac support is planned but is best-effort since macOS 27 drops support for those platforms. Currently the wheels ship with macOS deployment target 14.0 (the llvm-wheel toolchain's own floor). Linux aarch64 is planned as well.

The goal is to land the current wheel-building infrastructure before extending to more platforms, instead of having the entire thing arrive in one PR. That gives me a chance to verify the system before incrementally extending for more architectures (and more Python versions).

Regarding homebrew for eigen/boost: that's a test-time dependency, not part of how the wheel is built. The suite's probes now cover /opt/homebrew (arm64 brew) and /opt/local (MacPorts) as of the last commit, so whichever package manager provides eigen/boost, the tests find them.

On /opt/llvm: We build the llvm-wheel toolchain from compiler-research/ci-workflows on a manylinux image and OS X runner. Those archives are published on the recipe cache release (https://github.com/compiler-research/ci-workflows/releases/tag/cache) and this toolchain is staged at /opt/llvm by the 'before-all' step on linux and by the setup-recipe step in wheels.yml on macOS. I am still trying to figure out how to optimise this process further so bear that in mind.

.github/cibw_before_all_linux.sh
aaronj0: Only different in the fact that it is built on top of the manylinux pypa container to ensure it satisfies the glibc floor and "standard" build environment when building wheels. We then vendor it the same way. We will need to have a separate llvm-wheel because we can only support one LLVM version and build config per-wheel. Some projects just do it from scratch with an alma8 container such as Triton which we could adopt but this approach is more reliable since PyPA controls the image there which is well tested against Python package deployment environments.

vgvassilev: Does that mean that the other vendored llvms do not satisfy the libc versions?

aaronj0: No because they are source built on ubuntu24 runners. We need to build on a manylinux image to publish the wheels for manylinux. Similarly we will need a new image for manylinux-aarch64

pyproject.toml
```diff
@@ -30,6 +30,33 @@ provider = "scikit_build_core.metadata.regex"
...
+build = ["cp312-*", "cp313-*", "cp314-*"]
+skip = ["*-musllinux*"]
+build-verbosity = 1
+# Every wheel gets a smoke import; the full suite runs where the test
+# environment is a real runner: macOS below, linux in the test-wheel
+# job (the manylinux container's own CPython aborts mid-suite).
+test-sources = ["test", "requirements.txt", ".github/wheel_smoke.py"]
+test-command = "python .github/wheel_smoke.py"
+
+[tool.cibuildwheel.linux]
```
aaronj0, line 41: @vgvassilev You can see here that the vendored LLVM is reused exactly as build args on the machine that builds the wheel. Those machines are configured in the matrix in wheels.yml:

```
- { os: ubuntu-24.04, label: manylinux-x86_64, arch: x86_64 }
- { os: macos-26,     label: macosx-arm64,     arch: arm64 }
```

So for every new target platform we can add a new row and extend the wheels infra. I hope this addresses your concern of any redundancy here

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,139 @@
...
+        include:
+          - { os: ubuntu-24.04, label: manylinux-x86_64, arch: x86_64 }
+          - { os: macos-26,     label: macosx-arm64,     arch: arm64 }
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@v7
+
+      # ref pins the recipe content the cache key is computed from.
+      - uses: compiler-research/ci-workflows/actions/setup-recipe@main
```
aaronj0, line 46: Here we reuse the recipe we created with manylinux

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,139 @@
...
+      - uses: actions/upload-artifact@v7
+        with:
+          name: wheels-${{ matrix.label }}
+          path: wheelhouse/*.whl
+
+  sdist:
+    name: sdist
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v7
```
aaronj0, line 71: These are standard PyPA cibuildwheel and sdist actions required to generate the Python wheel

.github/cibw_before_all_linux.sh
vgvassilev: hm...

vgvassilev (commented):
I do not think I am competent enough to review this, but lgtm... @henryiii, do you mind having a look?

guitargeek (approved):
LGTM, thanks!

aaronj0:
As discussed, we can go ahead with this preliminary infrastructure to begin publishing wheels and follow up with extending to Linux aarch64 and intel-mac (best-effort).

PR #23, opened 2026-08-18
Drop stale/unreachable code and gCling. NFC

aaronj0:
Drops commented-out blocks, unused globals, py2-only paths, unused/hardcoded build-tree include paths and the emulated `cling::runtime::gCling` void pointer. NFC


Vipul-Cariappa (approved):
LGTM!

PR #24, opened 2026-08-18
[interop] Resolve libclangCppInterOp relative to libcppjit at runtime

aaronj0:
The backend used to locate `libclangCppInterOp` through absolute build-time paths, so a wheel built on one machine could not load on another.

This PR implements the following:
  - Library and headers are resolved relative to libcppjit's own load location, with the build-time install prefix as the fallback for build trees. This is necessary for the wheels building infrastructure (#22).
  
Validation: 
- full test suite passes locally across different cxx standard builds
- wheel relocatability: a wheel built in an isolated environment installs and runs in a second fresh venv after the build environment is deleted
- the install-prefix fallback verified against a bare build-directory `libcppjit.so`
- Ran clang-format-18

### DIFF
--- a/src/interop/interop_wrapper.cxx
+++ b/src/interop/interop_wrapper.cxx
@@ -23,6 +23,7 @@ using namespace cppjit;
 #include <csignal>
 #include <cstdlib> // for getenv
 #include <cstring>
+#include <filesystem>
 #include <iostream>
 #include <map>
 #include <mutex>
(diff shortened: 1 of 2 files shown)


src/interop/interop_wrapper.cxx
```diff
@@ -74,13 +75,37 @@ static inline bool is_integral(std::string& s) {
...
 
+struct InterOpPaths {
+  std::string Library;
+  std::string IncludeDir;
+};
+
+// One relative layout, two anchors: prefer CppInterOp next to our own load
+// location so wheels relocate; fall back to the build-time install prefix.
+static InterOpPaths cppinterop_paths() {
+  std::filesystem::path anchor = CPPINTEROP_INSTALL_PREFIX;
```
vgvassilev, line 86: We should probably try to detect an expected layout -- eg if the `.so` file is next to the executable/loader. In addition, the AddIncludePath call can happen at dlopen by the CppInterOp itself, if it knows better how to find back its include dir.

Vipul-Cariappa: On the second part of the comment, `dlopen`ing CppInterOp will never call `AddIncludePath` because there would be no interpreter created at that point.

Vipul-Cariappa (approved):
LGTM! Modulo Vassil's comment.

src/interop/interop_wrapper.cxx
vgvassilev: That's right, but we can probably hook in a callback that does it for us if that helps...

aaronj0: But at dlopen time, we still don't haven't loaded the dispatch table with the address of `AddIncludePath` and also no active interpreter so how would that work? Is the callback in CppInterOp to export its own include paths? And the callback must know the params/requirements of the interpreter that cppjit is trying to create.

aaronj0: >  it knows better how to find back its include dir.

We are already relying on CppInterOp's own build info for its include directory path. This mechanism controls the layout of how it's installed under `site-packages/cppjit` and uses the propagated variable:

```
  site-packages/
  ├── cppjit/                         # the pure-python package
  │   ├── __init__.py  _cpython_cppjit.py  _pythonization.py  _typemap.py
  │   ├── types.py  numba_ext.py  interactive.py  ll.py  reflex.py
  │   └── _stdcpp_fix.py  _version.py
  ├── libcppjit.so                    # 1.5 MB CPython extension
  ├── cppjit_backend/
  │   ├── __init__.py  _version.py    
  │   ├── lib/
  │   │   ├── libclangCppInterOp.so       # 247 MB — CppInterOp with static LLVM inside
  │   └── include/
  │       ├── CppInterOp/             # CppInterOp.h, CppInterOpTypes.h, Box.h, …
  │       ├── clang/Interpreter/      # clang API subset
  │       └── cpyrt/                  # the dispatcher API headers
  └── cppjit.libs/                    # auditwheel grafts: libzstd-*.so.1.5.7 (1.2 MB), libxml2, …
```

That effectively acts as what you mentioned: `detect an expected layout -- eg if the .so file is next to the executable/loader.`

vgvassilev: Ok, if it works with a simple expected layout as a first trial we can move forward and postpone the callback discussions.

aaronj0: @vgvassilev Okay if the rest looks good to you, can you approve?

vgvassilev (approved):

PR #25, opened 2026-08-18
[python] Anchor the cpyrt header probe on the extension module

aaronj0:
In editable installs the cpyrt header probe checked `__file__`, which maps to the source tree, so it missed the headers installed by scikit-build-core into site-packages and warned on every import. Now use the libcppjit extension's location, which resolves editable and regular installs identically. Also removes all dead lookups inherited with the probe (the `cpyrt` pip-distribution query via pkg_resources and the site/pythonX.Y layout guesses). `CPPJIT_API_PATH` still overrides and `"none"` still disables.

### DIFF
--- a/python/cppjit/__init__.py
+++ b/python/cppjit/__init__.py
@@ -58,19 +58,18 @@ import sysconfig
 import warnings
 
 try:
-    import __pypy__
+    import __pypy__  # noqa: F401
 
-    del __pypy__
-    ispypy = True
+    _ispypy = True
 except ImportError:
-    ispypy = False
+    _ispypy = False
 
 from . import _typemap
 from ._version import __version__ as __version__
 
 # import separately instead of in the above try/except block for easier to
 # understand tracebacks
-if ispypy:
+if _ispypy:
     raise ImportError("cppjit requires CPython; PyPy is not supported")
 from ._cpython_cppjit import *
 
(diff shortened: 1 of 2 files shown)


python/cppjit/__init__.py
```diff
@@ -334,82 +335,30 @@ def add_library_path(path):
...
-        ldversion = sysconfig.get_config_var("LDVERSION")
-        if not ldversion:
-            ldversion = sys.version[:3]
+    else:
+        _spec = importlib.util.find_spec("libcppjit")
+        if _spec is not None and _spec.origin:
+            apipath_extra = os.path.join(
+                os.path.dirname(_spec.origin), "cppjit_backend", "include"
+            )
+        del _spec
```
Vipul-Cariappa:
```suggestion
```

Why do we need to explicitly delete here? Wouldn't the GC collect it if it doesn't have any active references?

aaronj0: This is in module scope, the GC will only collect `_spec` once nothing references it, but here the module itself keeps a reference for the process lifetime (the assignment retains it as `cppjit._spec`), and it shows up in `dir(cppjit)`. The del drops that. We already do  `del` on  `apipath_extra, include_path, apipath, ispypy` for the same reason (also only required at init time and not a part of `dir(cppjit)`)

aaronj0: This has been addressed differently by moving all temp variables into function-locals (NFC change, second commit)

guitargeek (approved):
LGTM, thanks for the solid cleanup!

PR #29, opened 2026-08-19
[TEMP] Run only the nightly arm valgrind cell for iteration

aaronj0:
Attempt to fix the nightlies. The only remaining ARM vg nightlies fail because of a bug **in** valgrind:

It is deterministic: the same "Unrecognised instruction at 0x35b8702c" on every nightly run. The instruction is 0x38BFC108 = LDAPRB, an ARMv8.3 FEAT_LRCPC load-acquire byte, i.e. an atomic acquire load. The JIT emits it because the arm runner's CPU supports LRCPC, but Ubuntu 24.04's valgrind 3.22.0 can't decode it, so valgrind raises SIGILL and kills pytest mid-suite. Memcheck itself reports 0 errors. The last failure was `test_concurrent.py` addressed in #19  

Bug: https://bugs.kde.org/show_bug.cgi?id=476465
Fix: https://sourceware.org/git/?p=valgrind.git;a=commit;h=41e2f95cf129191555e0048ccbbb392ee0fb155e

But as mentioned that release that has the fix is not published for the ubuntu version in the CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,20 +18,14 @@ jobs:
     name: ${{ matrix.os }}/llvm${{ matrix.llvm }}/py${{ matrix.py }}/c++${{ matrix.cxx }}${{ matrix.vg && '/vg' || '' }}${{ matrix.flavor == 'cling' && '/cling' || '' }}
     strategy:
       fail-fast: false
-      # Mirrors CppInterOp's cppyy PR cells: clang-repl on the same OS/arch/LLVM
-      # /valgrind combos at Python 3.14, plus a cling cell and a C++23 cell
-      # (test_cpp23features runs only there). Breadth (3.12/3.13, C++17, LLVM
-      # 20, arm) lives in nightly.
+      # Iteration scaffold, not for merge: run the nightly arm valgrind
+      # cell (deterministic SIGILL, issue #9) plus one x86 valgrind cell
+      # on every push.
       matrix:
         include:
-          - { os: ubuntu-24.04,   llvm: '21', flavor: system, py: '3.14', cxx: '20', vg: true }
-          - { os: ubuntu-24.04,   llvm: '22', flavor: '',     py: '3.14', cxx: '20' }
-          - { os: ubuntu-24.04,   llvm: '22', flavor: '',     py: '3.14', cxx: '23' }
-          - { os: macos-26,       llvm: '21', flavor: system, py: '3.14', cxx: '20' }
-          - { os: macos-26-intel, llvm: '21', flavor: system, py: '3.14', cxx: '20' }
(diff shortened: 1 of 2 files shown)


mcbarton:
@aaronj0 @vgvassilev if you need a newer version of Valgrind than Ubuntu 24.04 offers due to this bug, then you can get the latest version  via conda/mamba like xeus-cpp does https://github.com/compiler-research/xeus-cpp/blob/82f9064929138f60995bfa12d0b82a382ff6a1b0/.github/workflows/main.yml#L194

aaronj0:
> @aaronj0 @vgvassilev if you need a newer version of Valgrind than Ubuntu 24.04 offers due to this bug, then you can get the latest version via conda/mamba like xeus-cpp does https://github.com/compiler-research/xeus-cpp/blob/82f9064929138f60995bfa12d0b82a382ff6a1b0/.github/workflows/main.yml#L194

Oh nice that's pretty useful, thanks.

test/conftest.py
```diff
@@ -0,0 +1,14 @@
...
+
+# valgrind (3.22 on the arm runners) cannot decode the FEAT_LRCPC loads
+# (LDAPR) the JIT emits under -march=native, and baseline armv8-a makes
+# clang outline atomics into __aarch64_* helper calls the JIT cannot
+# resolve; armv8.1-a keeps inline LSE atomics valgrind understands while
+# dropping LRCPC. Must happen before the first cppjit import boots the
+# interpreter.
+if os.getenv("IS_VALGRIND") and platform.machine() == "aarch64":
+    _extra = os.environ.get("CPPINTEROP_EXTRA_INTERPRETER_ARGS", "")
+    os.environ["CPPINTEROP_EXTRA_INTERPRETER_ARGS"] = (
```
mcbarton: Getting rid of an effective `-march=native` flag to fix the illegal instruction error is the same approach taken in a xeus-cpp PR previously. xeus-cpp had to remove this flag also in order to run Valgrind on arm without getting an illegal instruction error (see https://github.com/compiler-research/xeus-cpp/pull/449). I don't remember if it was the same illegal instruction error that this PR is trying to deal with. All I do remember is that the illegal instruction error didn't exist when I ran the tests on a Ubuntu arm system locally with the same os version locally. The illegal instruction error only ever occurred on the Github runner.

aaronj0: Here we are looking at what the JIT emits, which is not an illegal instruction but valgrind being spoofed into thinking it is one. I am using the flag only to test it but the fix is not to drop the flag. In the xeus-cpp case you point to, the tests should just run with march-native under vg

aaronj0:
This PR temporarily picks up the workflow built by https://github.com/compiler-research/ci-workflows/pull/101 which proves ti fix it. We can now run cppjit tests without suppressing on on vg + arm. Thanks for the conda suggestion @mcbarton

PR #31, opened 2026-08-20
[cmake] Require only Python Development.Module for the extension build

aaronj0:
Depend only on a subset of the Python Development component. We only require `Development.Module` when building our `libcppjit` extension, and in fact, manylinux images do not ship `Development.Embed` (`libpython`) so this solves that problem too.


guitargeek (approved):
Correct, thanks!

PR #32, opened 2026-08-20
[interop] Detect the resource dir of the required system LLVM

aaronj0:
As discussed in the last meeting, this is the first step towards dropping LLVM as a dependency (required for the wheels and binary distributions) and ship the required clang headers. We need to be able to reliably detect the resource dir. Currently `CreateInterpreter` picks up _any_ clang resource dir (naive `clang` probe) that will most likely _not_ be the one that cppjit's `libClangCppInterOp` was built against

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,11 +116,13 @@ add_library(cppjit SHARED ${CPYRT_SOURCES} ${INTEROP_SOURCES})
 add_dependencies(cppjit CppInterOp)
 
 # The wrapper anchors these relative spellings at its own load location,
-# falling back to the install prefix (see cppinterop_paths()).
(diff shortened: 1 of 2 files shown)


vgvassilev (approved):
We need to develop some low cost "packaging" ci where we test for the common configurations; and also we make sure the tests run successfully on the package-installed system...

aaronj0:
> We need to develop some low cost "packaging" ci where we test for the common configurations; and also we make sure the tests run successfully on the package-installed system...

Something along those lines are done in #22, where a relocatable binary is shipped to arbitrary images with common configurations (manylinux, osx) and we reproduce whether installing the package and  running tests pass. It's done in two steps, one host platform that builds the wheel and a target configuration that tests it.

I'll make sure to broaden the configs and keep the build cost low

PR #33, opened 2026-08-20
[build] Ship the clang builtin headers and prefer them at runtime

aaronj0:
This is the second part of the patch that ships the required runtime clang headers. With this change, a build/install of cppjit is self-contained and with the LLVM that is statically linked into `libClangCppInterOp.so`, we drop the LLVM dependency on a target machine. The wheels build PR will exercise this on clean images without LLVM.

As settled in review: the bundle is produced by the CMake install rules from the LLVM the build was pointed at, so it is identical to the headers `libClangCppInterOp` was compiled against under every install channel (pip, conda recipes, distro tooling, bare `cmake --install`) and for development builds against any `LLVM_DIR`. Only `include/` ships; the JIT reads nothing else from a resource directory. Setups without a bundle such as raw build trees, or distributions that strip vendored copies, fall back to the existing CppInterOp `DetectResourceDir` probing, which was the previous default behaviour (unchanged). This increases wheel size by 3-5MB

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,6 +125,7 @@ target_compile_definitions(cppjit PRIVATE
     CPPINTEROP_LIBRARY="cppjit_backend/lib/libclangCppInterOp${CMAKE_SHARED_LIBRARY_SUFFIX}"
     CPPINTEROP_INCLUDE_DIR="cppjit_backend/include"
     CPPJIT_CLANG_MAJOR="${LLVM_VERSION_MAJOR}"
+    CPPJIT_CLANG_INCLUDE_DIR="cppjit_backend/lib/clang/${LLVM_VERSION_MAJOR}"
 )
 
 target_include_directories(cppjit PRIVATE
@@ -174,6 +175,18 @@ install(CODE "
 install(CODE "
     file(INSTALL \"${CPPINTEROP_INSTALL_DIR}/include/\" DESTINATION \${CMAKE_INSTALL_PREFIX}/cppjit_backend/include)
 ")
+
+# ship the builtin headers of the build clang, laid out as a headers-only
+# resource dir: only include/ ships
+set(_clang_resource_dir "${LLVM_LIBRARY_DIR}/clang/${LLVM_VERSION_MAJOR}")
+if(NOT EXISTS "${_clang_resource_dir}/include")
+    message(FATAL_ERROR
+        "No builtin headers at ${_clang_resource_dir}/include; the LLVM at "
+        "${LLVM_DIR} carries no clang resource directory")
+endif()
+install(DIRECTORY "${_clang_resource_dir}/include/"
+    DESTINATION "cppjit_backend/lib/clang/${LLVM_VERSION_MAJOR}/include"
+)
 
 # the public cpyrt API headers keep their installed cpyrt/ prefix
 install(FILES
--- a/src/interop/interop_wrapper.cxx
+++ b/src/interop/interop_wrapper.cxx
@@ -78,6 +78,7 @@ static inline bool is_integral(std::string& s) {
 struct InterOpPaths {
   std::string Library;
   std::string IncludeDir;
+  std::string ClangIncludeDir; // empty when the bundled headers are absent
 };
 
 // One relative layout, two anchors: prefer CppInterOp next to our own load
@@ -94,8 +95,17 @@ static InterOpPaths cppinterop_paths() {
       anchor = here;
   }
 #endif
-  return {(anchor / CPPINTEROP_LIBRARY).string(),
-          (anchor / CPPINTEROP_INCLUDE_DIR).string()};
+  InterOpPaths Paths{(anchor / CPPINTEROP_LIBRARY).string(),
+                     (anchor / CPPINTEROP_INCLUDE_DIR).string(),
+                     {}};
+  // The builtin headers of the build clang ship with every installed
+  // package (see the CMake install rule); a raw build tree has none and
+  // falls back to resource-dir detection.
+  const std::filesystem::path bundled = anchor / CPPJIT_CLANG_INCLUDE_DIR;
+  std::error_code ec;
+  if (std::filesystem::exists(bundled / "include", ec))
+    Paths.ClangIncludeDir = bundled.string();
+  return Paths;
 }
 
 // The one place libclangCppInterOp is dlopen'd.
@@ -109,7 +119,8 @@ static bool loadDispatchAPI(const InterOpPaths& Paths) {
 
 // CppInterOp itself appends CPPINTEROP_EXTRA_INTERPRETER_ARGS inside
 // CreateInterpreter, so nothing needs to be forwarded from here.
-static interop::TInterp_t acquireOrCreateInterpreter() {
+static interop::TInterp_t
+acquireOrCreateInterpreter(const InterOpPaths& Paths) {
   if (auto existingInterp = Cpp::GetInterpreter())
     return existingInterp;
 
@@ -119,10 +130,11 @@ static interop::TInterp_t acquireOrCreateInterpreter() {
   args.push_back("-march=native");
 #endif
   // Without clang's builtin headers the interpreter fails at its first
-  // #include. CppInterOp probes only bare `clang`; when just
-  // clang-<major> is installed, resolve and pass it explicitly.
-  std::string resourceDir;
-  if (Cpp::DetectResourceDir("clang").empty())
+  // #include. Prefer the bundled copy: it matches the build clang and
+  // needs no LLVM on the host. DetectResourceDir refuses version
+  // mismatches, and CppInterOp itself probes only bare `clang`.
+  std::string resourceDir = Paths.ClangIncludeDir;
+  if (resourceDir.empty())
     resourceDir = Cpp::DetectResourceDir("clang-" CPPJIT_CLANG_MAJOR);
   if (!resourceDir.empty()) {
     args.push_back("-resource-dir");
@@ -219,7 +231,7 @@ extern "C" int LoadCppInterOp() {
     if (!loadDispatchAPI(Paths))
       return;
 
-    acquireOrCreateInterpreter();
+    acquireOrCreateInterpreter(Paths);
     configureInterpreter(Paths);
     preloadHeaders();
     defineRuntimeHelpers();


CMakeLists.txt
```diff
@@ -175,6 +176,19 @@ install(CODE "
...
+# requirement (the runtime probe prefers this copy; see cppinterop_paths())
+set(_clang_resource_dir "${LLVM_LIBRARY_DIR}/clang/${LLVM_VERSION_MAJOR}")
+if(NOT EXISTS "${_clang_resource_dir}/include")
+    message(FATAL_ERROR
+        "No builtin headers at ${_clang_resource_dir}/include; the LLVM at "
+        "${LLVM_DIR} carries no clang resource directory")
+endif()
+install(DIRECTORY "${_clang_resource_dir}/include/"
+    DESTINATION "cppjit_backend/lib/clang/${LLVM_VERSION_MAJOR}/include"
+)
```
Vipul-Cariappa, line 189: I suspect this is wrong. The clang resource directory contains a lot more than clang headers. Look at the contents of `clang -print-resource-dir`.

aaronj0: Yes but we only ship `/include`. As per my understanding that is all we need, and the in-process JIT does not use any symbols from the compiler-rt archives in `/lib`

aaronj0: The rest of the contents of resource-dir are not a requirement for a self-contained cppjit wheel. They are link time contents so they cannot be used unless we want to load those archives at runtime, all the necesseray llvm libs are already linked into cppinterop_

Vipul-Cariappa: If so, you may want to put it as include path? And not as resource directory?

aaronj0: Nothing in the installed layout is called resource director, but here it is written out as `clang/${LLVM_VERSION_MAJOR}/include` under `cppjit_backend`. Yes, the variables `CPPJIT_RESOURCE_DIR` and `ResourceDir` are inaccurate. I have now renamed them to say clang include dir.  `_clang_resource_dir` is still correct

aaronj0: ping @Vipul-Cariappa I believe your comment has been addressed.

src/interop/interop_wrapper.cxx
```diff
@@ -119,10 +130,10 @@ static interop::TInterp_t acquireOrCreateInterpreter() {
...
 #endif
   // Without clang's builtin headers the interpreter fails at its first
-  // #include. CppInterOp probes only bare `clang`; when just
-  // clang-<major> is installed, resolve and pass it explicitly.
-  std::string resourceDir;
-  if (Cpp::DetectResourceDir("clang").empty())
+  // #include. Prefer the bundled copy: it matches the build clang and
+  // needs no LLVM on the host.
+  std::string resourceDir = Paths.ClangIncludeDir;
+  if (resourceDir.empty() && Cpp::DetectResourceDir("clang").empty())
```
Vipul-Cariappa: Can we prioritize the system's resource directory.
First, check `clang-version`, then the one we include, then simple `clang`.

aaronj0: No that would not work. If we build a binary for manylinux or OS X based on llvm22 and a user installs it on their system which happens to have a lower LLVM version (apple always has clang) that would crash the interpreter. The point is that cppjit should not depend on system LLVM.

aaronj0: That is exactly the problem I described in last weeks meeting.

vgvassilev: > No that would not work. If we build a binary for manylinux or OS X based on llvm22 and a user installs it on their system which happens to have a lower LLVM version (apple always has clang) that would crash the interpreter. The point is that cppjit should not depend on system LLVM.

That would fail the match of the resource dir folder name which incorporates the version. We can also protect against older clangs such a check...

aaronj0: > That would fail the match of the resource dir folder name which incorporates the version. We can also protect against older clangs such a check...

Can you point me to which match fail you are referring to? Yes we can incorporate a check if the major version matches but just to clarify, we decided to bundle the headers so that we don't run into this problem and introduce mechanisms to ensure that system or package manager LLVM's agree with what cppjit needs.

If we always prioritize system-installed LLVM, that goes against the very solution you proposed (and we agreed on) to bundle the headers, leading to a self contained binary and not depend on an arbitrary LLVM picked up at runtime which can differ based on user environments

aaronj0: And that also seems to defeat the point, why should we pick up system or package manager LLVM when our own package installation has the same headers that `libClangCppInterOp` was built against.

Or vice-versa, if it is indeed the users responsibility to ensure that their system has the right LLVM version to support cppjit, then why should we bundle the headers at all, at the cost of increasing the wheel size

vgvassilev: > > That would fail the match of the resource dir folder name which incorporates the version. We can also protect against older clangs such a check...
> 
> Can you point me to which match fail you are referring to? Yes we can incorporate a check if the major version matches but just to clarify, we decided to bundle the headers so that we don't run into this problem and introduce mechanisms to ensure that system or package manager LLVM's agree with what cppjit needs.

The current packaging system is 1 packaging system -- the development mode still needs to work.

> If we always prioritize system-installed LLVM, that goes against the very solution you proposed (and we agreed on) to bundle the headers, leading to a self contained binary and not depend on an arbitrary LLVM picked up at runtime which can differ based on user environments

The bundling of the header files is a packaging decision, yes. However, outside of that package we need to make things flexible and work with compatible setups before we figure out how to deliver that within CppInterOp.

aaronj0: > The current packaging system is 1 packaging system -- the development mode still needs to work.

> The bundling of the header files is a packaging decision, yes. However, outside of that package we need to make things flexible and work with compatible setups before we figure out how to deliver that within CppInterOp.

It already works. Any LLVM that you build cppjit against, those headers will be used:

```
  pip install --no-build-isolation -ve . \
      --config-settings=build-dir=build-debug \
      --config-settings=cmake.build-type=Debug \
      --config-settings=cmake.define.LLVM_DIR=$LLVM_BUILD_DIR \
```

Here you can pass system LLVM or conda LLVM or your own source build and this logic is still robust against that because we will pick up those exact headers that `libClangCppInterOp` was built against.

aaronj0: What the review suggestion here proposes is to prioritize the arbitrary clang headers detected at runtime over the one that we bundle and I disagree with that. If anything, system clang can be used as a fallback but not as the first pick for resource dir since there is no guarantee that exists or even work if it does.

vgvassilev: > pip install --no-build-isolation -ve .

That's 1 package manager. Does that solution work for non-pip setups?

> What the review suggestion here proposes is to prioritize the arbitrary clang headers detected at runtime over the one that we bundle and I disagree with that. If anything, system clang can be used as a fallback but not as the first pick for resource dir since there is no guarantee that exists or even work if it does.

Where my comments suggested that?

aaronj0: > That's 1 package manager. Does that solution work for non-pip setups?

Yes. Development mode works because this is not a pip-specific solution but in the the underlying build system. The installation layout is fixed and under our control irrespective of the package manager. Here is a pure CMake build, without pip, against an uninstalled developer build tree of LLVM:

```console
$ python3 -m venv demo-venv          # python to build against
$ cmake -S cppjit -B build \
      -DCMAKE_BUILD_TYPE=Release \
      -DLLVM_DIR=$HOME/llvm-project/release_build/lib/cmake/llvm \
      -DPython_EXECUTABLE=$PWD/demo-venv/bin/python
$ cmake --build build -j$(nproc)
$ cmake --install build --prefix $PWD/cppjit-install
$ cp -r cppjit/python/cppjit cppjit-install/   # normally done by the wheel
```

The installed tree carries the builtin headers of that LLVM:

```console
$ diff -r $HOME/llvm-project/release_build/lib/clang/21/include \
          cppjit-install/cppjit_backend/lib/clang/21/include && echo IDENTICAL
IDENTICAL
```

So irrespective of what LLVM we develop against, the runtime prefers the headers `libclangCppInterOp` was compiled against. In a developer's uninstalled build tree there are no bundled headers, and the loader falls back to `DetectResourceDir`, which is unchanged in this patch.

The result is self-contained either way. With an empty `PATH` (no clang visible to the process):

```console
$ PYTHONPATH=$PWD/cppjit-install PATH=/nonexistent demo-venv/bin/python \
      -c "import cppjit; cppjit.cppdef('#include <vector>\nint f(){ std::vector<int> v{41}; return v[0]+1; }'); print('f() =', cppjit.gbl.f())"
f() = 42
```

The point: This does not depend on a package manager, but a property of the CMake install rules which is run by every install channel. At the end of the day, cppjit is a Python package and any PEP 517 (Python standard that separates the build frontend from the build backend) frontend, drives it through scikit-build-core. So this satisfies conda recipes, conda build environments, a linux distro specific package manager, and a bare `cmake --install` in a developers setup (with no Python packaging at all). We are satisfying the Python build standard. If any distribution or install does not contain the bundled headers (which is impossible unless someone removes them on purpose) we try the `DetectResourceDir` fallback with `clang`. We can improve that fallback to restrict to the supported LLVM version but that is outside the scope of this PR.

aaronj0: > Where my comments suggested that?

I was referring to the original review comment in this thread:

> Can we prioritize the system's resource directory.
> First, check clang-version, then the one we include, then simple clang.

cc @Vipul-Cariappa

vgvassilev: > > That's 1 package manager. Does that solution work for non-pip setups?
> 
> Yes. Development mode works because this is not a pip-specific solution but in the the underlying build system. The installation layout is fixed and under our control irrespective of the package manager. Here is a pure CMake build, without pip, against an uninstalled developer build tree of LLVM:
> 
> ```
> $ python3 -m venv demo-venv          # python to build against
> $ cmake -S cppjit -B build \
>       -DCMAKE_BUILD_TYPE=Release \
>       -DLLVM_DIR=$HOME/llvm-project/release_build/lib/cmake/llvm \
>       -DPython_EXECUTABLE=$PWD/demo-venv/bin/python
> $ cmake --build build -j$(nproc)
> $ cmake --install build --prefix $PWD/cppjit-install
> $ cp -r cppjit/python/cppjit cppjit-install/   # normally done by the wheel
> ```
> 
> The installed tree carries the builtin headers of that LLVM:
> 
> ```
> $ diff -r $HOME/llvm-project/release_build/lib/clang/21/include \
>           cppjit-install/cppjit_backend/lib/clang/21/include && echo IDENTICAL
> IDENTICAL
> ```
> 
> So irrespective of what LLVM we develop against, the runtime prefers the headers `libclangCppInterOp` was compiled against. In a developer's uninstalled build tree there are no bundled headers, and the loader falls back to `DetectResourceDir`, which is unchanged in this patch.
> 
> The result is self-contained either way. With an empty `PATH` (no clang visible to the process):
> 
> ```
> $ PYTHONPATH=$PWD/cppjit-install PATH=/nonexistent demo-venv/bin/python \
>       -c "import cppjit; cppjit.cppdef('#include <vector>\nint f(){ std::vector<int> v{41}; return v[0]+1; }'); print('f() =', cppjit.gbl.f())"
> f() = 42
> ```
> 
> The point: This does not depend on a package manager, but a property of the CMake install rules which is run by every install channel. At the end of the day, cppjit is a Python package and any PEP 517 (Python standard that separates the build frontend from the build backend) frontend, drives it through scikit-build-core. So this satisfies conda recipes, conda build environments, a linux distro specific package manager, and a bare `cmake --install` in a developers setup (with no Python packaging at all). We are satisfying the Python build standard. If any distribution or install does not contain the bundled headers (which is impossible unless someone removes them on purpose) we try the `DetectResourceDir` fallback with `clang`. We can improve that fallback to restrict to the supported LLVM version but that is outside the scope of this PR.

Ok, that clarifies it -- can you update the PR description and the commit message? That would have helped me.

Vipul-Cariappa: What I am proposing here is that if the system LLVM version matches our LLVM version, we use the system's resource directory; otherwise, the one we ship. Similar to what is being done with the ROOT built-ins.
Look at the code you have now. You first check `std::string resourceDir = Paths.ClangIncludeDir; if (resourceDir.empty() ...`, then `Cpp::DetectResourceDir("clang")`, then `Cpp::DetectResourceDir("clang-" CPPJIT_CLANG_MAJOR)`.

> Can you point me to which match fail you are referring to? 

https://github.com/compiler-research/CppInterOp/blob/1cfe7ee0da5e7abeb80982b9e67270806884aa52/lib/CppInterOp/CppInterOp.cpp#L5419-L5421

aaronj0: > What I am proposing here is that if the system LLVM version matches our LLVM version, we use the system's resource directory; otherwise, the one we ship. Similar to what is being done with the ROOT built-ins. Look at the code you have now. You first check `std::string resourceDir = Paths.ClangIncludeDir; if (resourceDir.empty() ...`, then `Cpp::DetectResourceDir("clang")`, then `Cpp::DetectResourceDir("clang-" CPPJIT_CLANG_MAJOR)`.
> 

This is a different case as the ROOT builtins, no ROOT libraries built with some builtin library headers is going to then prefer the ones found on the system at runtime. (e.g A ROOT binary's lib A built with package B headers do not select system package B headers instead of the ones bundled in the binary). It only chooses to build the built-in version of the library if it does not find an existing installation (which is build-time). For the Python package (and any install), ROOT bundles all of the required clang headers, and the interpreter uses only those headers. This is for the same reason and does not prefer the ones that may exist on the system.

The logic you are talking about exists when running a source build of cppjit. If you clone the repo and try to pip install, we do `find_package`  to look for a compatible system LLVM to build the package. This saves disk space and build time since we don't need a package manager installation or externalproject_add. But once the package is built/installed with an LLVM source and we have already bundled the headers from the same source, you guarantee that it will always work.

With that approach, the benefit of preferring system headers is also non-existent. The bundled headers exist anyway, and we do not save build time (as this is a runtime probe) or disk space.
 
> https://github.com/compiler-research/CppInterOp/blob/1cfe7ee0da5e7abeb80982b9e67270806884aa52/lib/CppInterOp/CppInterOp.cpp#L5419-L5421

Based on the check you have pointed to, I only need to do `Cpp::DetectResourceDir("clang-" CPPJIT_CLANG_MAJOR)`. I can correct that, thanks.

Vipul-Cariappa: Could you check whether, when using the headers packages by CppJIT, you can run OpenMP code? If yes, then there is no problem with defaulting to what we provide. If not, we will need to think about what the correct default is. I am talking about the case where the system LLVM matches CppJIT's LLVM version.

aaronj0: Yes, with the current system, `omp.h` lands in the resource dir's `include/` when the LLVM build is openmp enabled.

Of course, you do need the openmp runtime which needs to be present on the system, which is standard behavior for third-party libraries. If I build a wheel and install it on a system with no LLVM, the OpenMP code still compiles and works. The wheel carries `omp.h` and runs as soon as the runtime is loaded.

Here's how I verified this: built a wheel from this branch against an openmp-enabled LLVM 21 which ships `cppjit_backend/lib/clang/21/include/omp.h`. I then installed it in a bare `ubuntu:24.04` container with only `g++`, `libomp-dev` and `python3` (no LLVM):

```python
import cppjit
cppjit.load_library("libomp.so.5")
cppjit.cppdef('''
#include <omp.h>
int maxthreads() { return omp_get_max_threads(); }
long parsum(int n) {
  long s = 0;
  #pragma omp parallel for reduction(+:s)
  for (int i = 1; i <= n; ++i) s += i;
  return s;
}
''')
print(cppjit.gbl.maxthreads())    # 16
print(cppjit.gbl.parsum(100000))  # 5000050000
```

Here we need to pass `-fopenmp` in `CPPINTEROP_EXTRA_INTERPRETER_ARGS` as usual. We can either LD_PRELOAD or load_library `libomp`

vgvassilev: Well, the openmp is a separate problem. The one we have is the one shipped with llvm and we need to be careful because last time I checked the intel implementation had some problems to be used. If the user tries to use the system omp (which gcc and clang allow for) we have no way to do that. I guess we need to make sure we have some level of diagnostics in that case.

aaronj0: > I guess we need to make sure we have some level of diagnostics in that case.

That makes sense, can be addressed in a follow-up.

vgvassilev (approved):
Lgtm!

PR #35, opened 2026-08-21
[memory-analysis] Added a field to CPPMethod to store memory-ownershi…

keremsahn:
When overload resolution happens and a suitable candidate found and called, HandleReturn helper is called for every case. In this function I call GetAllocBehaviour function for the first element of overload candidates (I followed design where IsCreator flag belongs to overload group, not unique for every PyCallable, but I think of changing it). 

@Vipul-Cariappa @aaronj0

### DIFF
--- a/src/cpyrt/CPPMethod.cxx
+++ b/src/cpyrt/CPPMethod.cxx
@@ -753,6 +753,14 @@ PyObject* cpyrt::CPPMethod::GetArgDefault(int iarg, bool silent) {
 bool cpyrt::CPPMethod::IsConst() { return interop::IsConstMethod(GetMethod()); }
 
 //----------------------------------------------------------------------------
+interop::AllocType cpyrt::CPPMethod::GetAllocBehaviour() {
+  if (fAllocType.has_value())
+    return *fAllocType;
+  interop::AllocType attrResult = interop::IsAllocator(GetMethod());
+  fAllocType = attrResult;
+  return attrResult;
+}
+//----------------------------------------------------------------------------
 PyObject* cpyrt::CPPMethod::GetScopeProxy() {
   // Get or build the scope of this method.
   return CreateScopeProxy(fScope);
--- a/src/cpyrt/CPPMethod.h
+++ b/src/cpyrt/CPPMethod.h
@@ -5,6 +5,7 @@
 #include "PyCallable.h"
 
 // Standard
+#include <optional>
 #include <string>
 #include <unordered_map>
 #include <vector>
@@ -62,6 +63,7 @@ public:
   PyObject* GetCoVarNames() override;
   PyObject* GetArgDefault(int iarg, bool silent = true) override;
   bool IsConst() override;
+  cppjit::interop::AllocType GetAllocBehaviour() override;
 
   PyObject* GetScopeProxy() override;
   interop::TCppFuncAddr_t GetFunctionAddress() override;
@@ -116,6 +118,7 @@ private:
 protected:
   // cached value that doubles as initialized flag (uninitialized if -1)
   int fArgsRequired;
+  std::optional<cppjit::interop::AllocType> fAllocType;
 };
 
 } // namespace cppjit::cpyrt
--- a/src/cpyrt/CPPOverload.cxx
+++ b/src/cpyrt/CPPOverload.cxx
@@ -155,6 +155,12 @@ static inline PyObject* HandleReturn(CPPOverload* pymeth, CPPInstance* im_self,
     CPPInstance* cppres =
         (CPPInstance*)(CPPInstance_Check(result) ? result : nullptr);
 
+    interop::AllocType AT =
+        pymeth->fMethodInfo->fMethods[0]->GetAllocBehaviour();
+    if (AT != interop::AllocType::None && AT != interop::AllocType::Null &&
+        AT != interop::AllocType::Unknown)
+      pymeth->fMethodInfo->fFlags |= CallContext::kIsCreator;
+
     // if this method creates new objects, always take ownership
     if (IsCreator(pymeth->fMethodInfo->fFlags)) {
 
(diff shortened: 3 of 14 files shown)


src/cpyrt/CPPMethod.cxx
```diff
@@ -749,6 +749,23 @@ PyObject* cpyrt::CPPMethod::GetArgDefault(int iarg, bool silent) {
 
 bool cpyrt::CPPMethod::IsConst() { return interop::IsConstMethod(GetMethod()); }
 
+//----------------------------------------------------------------------------
+// FIXME: For now every allocation is assumed to be done with `new`
+// will be fixed soon. Also the reason function returns an AllocType
+// instead of bool, IsAllocator function from CppInterOp changed a little,
```
aaronj0: You can test these changes on top of your CppInterOp branches by updating the remote and ref here: https://github.com/compiler-research/cppjit/blob/main/CMakeLists.txt#L14-L15

Vipul-Cariappa: The changes this PR depends on are already part of main in CppInterOp.

Vipul-Cariappa:
I am converting to draft for now. Once I am happy with the state of this PR, I will mark it "ready for review" for others.
Note: I may take some time to review it.
@keremsahn, hope that is fine with you.

test/cpp/memory_analysis.h
```diff
@@ -0,0 +1,13 @@
...
+#define MEMORY_ANALYSIS_H
+
+#include <new>
+#include <stdlib.h>
+
+class memAnalysisKlass {
+  int val;
+};
+__attribute__((malloc)) memAnalysisKlass* allocTest();
+__attribute__((ownership_returns(malloc))) memAnalysisKlass* allocTestReturns();
```
Vipul-Cariappa: Put everything in this file into a namespace; otherwise, the names may conflict with things that are JIT compiled or exist in other files.

test/test_memoryanalysis.py
```diff
@@ -0,0 +1,31 @@
...
+        obj = cppjit.gbl.allocTest()
+        assert type(obj) == cppjit.gbl.memAnalysisKlass
+        assert obj.__python_owns__
+
+    def test02_ownership_returns_attr(self):
+        import cppjit
+
+        obj = cppjit.gbl.allocTestReturns()
+        assert type(obj) == cppjit.gbl.memAnalysisKlass
+        assert obj.__python_owns__
```
Vipul-Cariappa, line 54: We also need tests for
```python
assert not obj.__python__owns__
```

src/cpyrt/PyCallable.h
```diff
@@ -37,6 +37,9 @@ class PyCallable {
   virtual PyObject* GetCoVarNames() = 0;
   virtual PyObject* GetArgDefault(int /* iarg */, bool silent = true) = 0;
   virtual bool IsConst() { return false; }
+  virtual cppjit::interop::AllocType GetAllocBehaviour() {
+    return cppjit::interop::AllocType::None;
```
Vipul-Cariappa, line 41:
```suggestion
    return cppjit::interop::AllocType::Unknown;
```
?

Vipul-Cariappa (commented):
Some minor comments. LGTM, otherwise.

src/cpyrt/PyCallable.h
keremsahn: Only class does not override GetAllocBehaviour is TPythonCallBack, and from my understanding this class is for some sort of python function, not C/C++, therefore I thought it would make sense to return None

src/cpyrt/CPPMethod.cxx
aaronj0: I meant for changes that have not been merged yet (referred to in the FIXME comment on `IsAllocator`). iiuc the API on the CppInterOp side will change pending some work being merged, so if you want to coverage test the cppjit patch, this could help

keremsahn:
@Vipul-Cariappa I applied Aaron's suggestion, this PR will fetch CppInterOp from my fork's latest branch, so I added bunch of things, worth reviewing again

aaronj0:

src/cpyrt/CPPMethod.h
```diff
@@ -62,6 +63,7 @@ class CPPMethod : public PyCallable {
   PyObject* GetCoVarNames() override;
   PyObject* GetArgDefault(int iarg, bool silent = true) override;
   bool IsConst() override;
+  cppjit::interop::AllocType GetAllocBehaviour() override;
```
Vipul-Cariappa, line 66:
```suggestion
  interop::AllocType GetAllocBehaviour() override;
```

src/cpyrt/CPPOverload.cxx
```diff
@@ -155,6 +155,12 @@ static inline PyObject* HandleReturn(CPPOverload* pymeth, CPPInstance* im_self,
     CPPInstance* cppres =
         (CPPInstance*)(CPPInstance_Check(result) ? result : nullptr);
 
+    interop::AllocType AT =
+        pymeth->fMethodInfo->fMethods[0]->GetAllocBehaviour();
+    if (AT != interop::AllocType::None && AT != interop::AllocType::Null &&
+        AT != interop::AllocType::Unknown)
+      pymeth->fMethodInfo->fFlags |= CallContext::kIsCreator;
+
```
Vipul-Cariappa, line 163: Can you change the function signature of `HandleReturn` to accept a `PyCallable`, and use it instead of assuming that `fMethods[0]` is the correct overload?

Vipul-Cariappa (commented):
A few changes:

Vipul-Cariappa:
Could you also please rebase?

src/cpyrt/CPPOverload.cxx
Vipul-Cariappa: Should we also add a check, `if (return_value_is_pointer) ...`?

keremsahn: > Can you change the function signature of `HandleReturn` to accept a `PyCallable`, and use it instead of assuming that `fMethods[0]` is the correct overload?

That was in my mind, to distinguish between overloads, but I thought this should be content of another PR since how overloads are stored and they share same flags, so there are work should be done outside of ownership.

keremsahn: > Should we also add a check, `if (return_value_is_pointer) ...`?

Yes, for now we should add it I guess, until records are handled properly

PR #37, opened 2026-08-22
[WIP] On demand analyze for allocation

keremsahn:
Built upon #35 
Added wrapping for GetAllocType, and it is called if user sets gUseAllocAnalyze variable, which is disabled by default. Analyzer has lower priority compared to attribute check.

A detail: Analyzer caches the result, so once the function is called with gUseAllocAnalyze, the result stays, even though the flag is set to False laterly.
@aaronj0 @vgvassilev @Vipul-Cariappa

### DIFF
--- a/test/cpp/memory_analysis.h
+++ b/test/cpp/memory_analysis.h
@@ -1,0 +1,36 @@
+#ifndef MEMORY_ANALYSIS_H
+#define MEMORY_ANALYSIS_H
+
+#include <new>
+#include <stdlib.h>
+namespace memory {
+
+class memAnalysisKlass {
+public:
+  int val;
+};
+__attribute__((malloc)) memAnalysisKlass* mallocAttr();
+__attribute__((ownership_returns(malloc))) memAnalysisKlass*
+ownershipReturnsAttr();
+memAnalysisKlass* noAttr();
+
+struct memOwn {
+  int val;
+  memOwn(int value) : val(value) {}
+  memOwn() { val = 0; }
+  // Attribute injected by APINotes
+  static memOwn* memOwnAllocator(int x) { return new memOwn(x); }
+};
+
+// Attribute injected by APINotes
+memOwn* memOwnAllocGlobal();
+
+// Attribute injected by redeclaration
+memOwn* allocDefaultMemOwn();
+
+// No ownership attribute anywhere
+memOwn* noAttrAlloc();
+
+} // namespace memory
+
+#endif // MEMORY_ANALYSIS_H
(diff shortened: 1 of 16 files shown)


test/test_memoryanalysis.py
```diff
@@ -29,3 +29,20 @@ def test02_ownership_returns_attr(self):
         obj = cppjit.gbl.allocTestReturns()
         assert type(obj) == cppjit.gbl.memAnalysisKlass
         assert obj.__python_owns__
+
+    def test03_analyzer_new(self):
+        import cppjit
+
+        cppjit._backend.SetUseAllocAnalyzer(True)
```
Vipul-Cariappa: Please add a wrapper in `__init__.py`, such that the user can directly do
`cppjit.use_alloc_analyzer(...)`. Note that in Python, function names should be in snake case.

test/test_memoryanalysis.py
```diff
@@ -29,3 +29,20 @@ def test02_ownership_returns_attr(self):
...
+        assert type(obj) == cppjit.gbl.memAnalysisKlass
+        assert obj.__python_owns__
+
+    def test04_analyzer_new_default(self):
+        import cppjit
+
+        cppjit._backend.SetUseAllocAnalyzer(False)
+        obj = cppjit.gbl.allocNew2()
+        assert type(obj) == cppjit.gbl.memAnalysisKlass
+        assert not (obj.__python_owns__)
```
Vipul-Cariappa: We need a test for reusing existing results... i.e. already analyzed, but `use_alloc_analyzer(False)`.

test/cpp/memory_analysis.h
```diff
@@ -10,4 +10,6 @@ class memAnalysisKlass {
 __attribute__((malloc)) memAnalysisKlass* allocTest();
 __attribute__((ownership_returns(malloc))) memAnalysisKlass* allocTestReturns();
 
+inline memAnalysisKlass* allocNew() { return new memAnalysisKlass; }
+inline memAnalysisKlass* allocNew2() { return new memAnalysisKlass; }
```
Vipul-Cariappa: Can the function names be a bit more descriptive? In the context of the test. Example:
`AnalysisOnAllocNew` & `AnalysisOffAllocNew`?

Vipul-Cariappa (commented):
Minor comments. LGTM!

test/cpp/memory_analysis.h
keremsahn: Changed

test/test_memoryanalysis.py
keremsahn: Added

test/test_memoryanalysis.py
keremsahn: Added the wrapper, it is cppjit.use_alloc_analyzer()

aaronj0:

Vipul-Cariappa:
These changes look good to me. If you can rebase this, will merge immediately after #35.

PR #40, opened 2026-08-25
Penalise void* arguments in overload priority as intended

aaronj0:
Based on the work from the compiler research forks: https://github.com/compiler-research/CPyCppyy/pull/225 and https://github.com/compiler-research/cppyy/pull/239
Migrated with AI

cc @conrade-ctc


conrade-ctc:
whoops, I think you meant to reference https://github.com/compiler-research/CPyCppyy/pull/225!

aaronj0:
> whoops, I think you meant to reference [compiler-research/CPyCppyy#225](https://github.com/compiler-research/CPyCppyy/pull/225)!

Ah, yes. Has been updated. The code itself is from the correct PR :)

PR #41, opened 2026-08-25
Improve error reporting for method calls without a C++ object

aaronj0:
Based on the work from the compiler research forks: https://github.com/compiler-research/CPyCppyy/pull/223 and https://github.com/compiler-research/cppyy/pull/237
Migrated with AI

cc @rybkine


rybkine:
@aaronj0 Thank you! Everything looks good.

vgvassilev:
Can we take the claude authorship outside of the commit log/loop?

aaronj0:
> Can we take the claude authorship outside of the commit log/loop?

Ah, not sure how we can do that now..

PR #42, opened 2026-08-26
Drop cppjit_backend and unify with cpyrt

aaronj0:
This PR unifies our installation layout under a single cppjit module, dropping the concept of a separate `cppjit_backend`, making our wheel package count 1. As a part of this change, I have dropped the long-standing header duplication (previously `Cppyy.h` and `cpp_cppyy.h`) between cpyrt and interop, so that there is only a single source of definitions for the `cppjit::interop` API and types.


guitargeek (approved):
Fantastic, thanks!

vgvassilev (approved):
Lgtm! Consider either splitting into two PRs or making once commit with the appropriate commit message.

Issue #47, opened 2026-08-27
Fails with ` pytest -n $(nproc --all)`


keremsahn:
1)
```python
def test09_templated_function(self):
        """Templated free function"""
    
        import cppjit
    
>       mul = cppjit.gbl.multiply
              ^^^^^^^^^^^^^^^^^^^
E       AttributeError: <namespace cppjit.gbl at 0x2647e8b0> has no attribute 'multiply'. Full details:
E         type object '' has no attribute 'multiply'
E         'multiply' is not a known C++ class
E         'multiply' is not a known C++ template
E         'multiply' is not a known C++ enum

test/test_doc_features.py:717: AttributeError
```

Reason: multiply is defined at `class TestDOCFEATURES`, but test09 is at `class TestTUTORIALFEATURES`

2) Same for `test_lowlevel.py::TestLOWLEVEL::test03_memory` :  `AttributeError: module 'cppjit' has no attribute 'll'`
The `ll` is not directly included at test03, and when test03 runs before one of the tests that import `ll` finishes, test fails

Vipul-Cariappa:
I suspect this is a pytest bug, or maybe it is expected behaviour of pytest, and we should update the test...
As I understand it, `setup_class` needs to run before the test itself, but when running tests individually, `setup_class` is not run. I guess when running in parallel, something similar happens. 
Can you also test my hypothesis about `setup_class`, and check pytest's documentation about it?

For `cppjit.ll`, can you add `import cppjit.ll` at the start of the test, or in `setup_class`? Check if that solves the problem.

aaronj0:
Fixed with #49

PR #49, opened 2026-08-30
[FIX]: Fails for parallel test

keremsahn:
I solved the causes described in #47 
@Vipul-Cariappa @aaronj0

### DIFF
--- a/test/test_doc_features.py
+++ b/test/test_doc_features.py
@@ -142,10 +142,6 @@ int call_int_int_function(int (*f)(int, int), int i1, int i2) {
     return f(i1, i2);
 }
 
-template<class A, class B, class C = A>
-C multiply(A a, B b) {
-    return static_cast<C>(a * b);
-}
 
 //-----
 namespace Namespace {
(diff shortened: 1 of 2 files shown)


aaronj0:
I think these changes are already a part of ongoing test-revamp in this branch: https://github.com/compiler-research/cppjit/tree/test-suite-infra, but we can land this independently. Thanks.

aaronj0 (approved):
Lgtm

vgvassilev:
Can you explain how this change resolved the parallelism problems?

aaronj0:
> Can you explain how this change resolved the parallelism problems?

`test09_templated_function` (TestTUTORIALFEATURES) relied on another class (TestDOCFEATURES) to cppdef `multiply` introducing an inter-test setup dependency.

`test03_memory` never imported ll itself making the module unreachable unless the same worker happened to have an earlier test that did so.

This change was already tested with https://github.com/compiler-research/cppjit/commit/87e799c3d8074e51a7294c1f4d20fe287c41b3a0

keremsahn:
> Can you explain how this change resolved the parallelism problems?

The first change was just a definition replacement, I just moved the definition of multiply function, where it is actually used.

In the second I just added import ll, the test was using ll but was not importing it, so it was dependent on previous tests' import, so it was failing sometimes in parallel mode

vgvassilev:
Ok, thanks -- in retrospect, probably worth being part of the commit message.

PR #50, opened 2026-08-30
[build] Stage CppInterOp and ship one stripped libclangCppInterOp

aaronj0:
Makes use of https://github.com/compiler-research/CppInterOp/pull/1097 for leaner installs dropping the wheel size by about half. Also updates to latest CppInterOp in the process.

### DIFF
--- a/cmake/AddCppInterOp.cmake
+++ b/cmake/AddCppInterOp.cmake
@@ -22,7 +22,9 @@ function(cppjit_add_cppinterop)
         -DLLVM_DIR=${LLVM_DIR}
         -DCPPINTEROP_ENABLE_TESTING=${CPPJIT_ENABLE_CPPINTEROP_TESTS}
         -DBUILD_SHARED_LIBS=ON
-        -DCMAKE_INSTALL_PREFIX=${CPPINTEROP_INSTALL_DIR}
+        # The wheel ships a single unversioned library file.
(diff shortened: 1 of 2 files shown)


cmake/AddCppInterOp.cmake
```diff
@@ -9,6 +9,9 @@ include(ExternalProject)
 option(CPPJIT_USE_CLING "Build CppInterOp against a prebuilt Cling C++ Interpreter (ROOT)" OFF)
 mark_as_advanced(CPPJIT_USE_CLING)
 
+# Stripping keeps .dynsym, so the dlsym-based dispatch still resolves.
+option(CPPJIT_STRIP_CPPINTEROP "Strip the staged libclangCppInterOp" ON)
```
vgvassilev: Why do we need a new option? Under what conditions we need the non-stripped version?

aaronj0: Yes, ideally we can just keep the stripped version always, but if a developer working on CppInterOp wants the entire install tree accessible by the python extension this would allow you to do so. Can drop it if you think it is not needed

vgvassilev: I think we should not require to install CppInterOp in a dev setup. It should work from the debug folder.

aaronj0: Addressed

vgvassilev (approved):
LGTM!

PR #51, opened 2026-08-31
[test] Improve pytest setup, markers and support for xdist

### LINKED ISSUE #46: Fix syntax errors in tests
If you inspect the ci you'll see quite a few tests which error due to a syntax error. For example

```
[779](https://github.com/compiler-research/cppjit/actions/runs/33078956981/job/98540584529?pr=45#step:8:1790)
    E     File "/home/runner/work/cppjit/cppjit/test/test_overloads.py", line 218
    E       @mark.skipif(condition=not not IS_MAC_ARM, condition=IS_MAC, reason="Seg Faults")
    E                                                  ^^^^^^^^^^^^^^^^
    E   SyntaxError: keyword argument repeated: condition
```

By fixing these syntax errors you might get some more passing tests, but at the very least you'll be able to test things you couldn't test before.

aaronj0:
Fixes a large set of issues in our pytest configurations, markers, support for multi-CPU execution with xdist.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@ __pycache__/
 
 # Built test dictionaries and extension modules
 *.so
+*.so.*.tmp
+*Dict.lock
 
 # Packaging
 build/
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,7 @@ input = "python/cppjit/_version.py"
 [tool.pytest.ini_options]
 testpaths = ["test"]
 pythonpath = ["test"]
+xfail_strict = true
 
 [tool.ruff]
 show-fixes = true
--- a/test/Makefile
+++ b/test/Makefile
@@ -29,8 +29,9 @@ ifeq ($(PLATFORM),Darwin)
   cppflags+=-dynamiclib -single_module -undefined dynamic_lookup -Wno-delete-non-virtual-dtor
 endif
 
-cpp/%Dict.so: cpp/%.cxx
-	$(CXX) $(cppflags) -shared -o $@ $^
+# a worker can load the library while another rebuilds it, so publish it whole
+cpp/%Dict.so: cpp/%.cxx cpp/%.h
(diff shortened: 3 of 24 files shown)


aaronj0:
The earlier CI failures were a result of `xfail-strict`. This treats XPASS as a fail, which is good as we no longer will have XPASSes. The latest commit now flips those. (mostly on Mac)

guitargeek (approved):
LGTM!

.github/workflows/ci.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           - { os: macos-26-intel, llvm: '21', flavor: system, py: '3.14', cxx: '20' }
           # cling backend, against the cached llvm-root (cling-llvm22) recipe cell.
           - { os: ubuntu-24.04,   llvm: '22', flavor: cling,  flavor_version: cling-llvm22, py: '3.14', cxx: '20' }
-    uses: compiler-research/ci-workflows/.github/workflows/cppjit.yml@main
+    uses: aaronj0/ci-workflows/.github/workflows/cppjit.yml@cppjit-marker-driven-sweep
```
vgvassilev: The ci-workflows/bin/repro has a `--ci-workflows` parameter which allows claude to iterate locally: https://github.com/compiler-research/ci-workflows/blob/main/docs/developer-guide.md#what---ci-workflows-path-does

aaronj0: Thanks. I do use bin/repro with the updated workflow locally before opening PRs but I need to do this here to ensure the results are good across the CI matrix before landing the changes on ci-workflows side. Otherwise I have to land the changes in ci-workflows before my PR in cppjit can actually use the new workflow. I just found the span issue on mac as a result of that.

vgvassilev: If you are running on mac and have docker then that's also possible locally.

aaronj0: Yes indeed, but I am running only on Linux

mcbarton:
can someone link this PR to closing https://github.com/compiler-research/cppjit/issues/46 since I think this PR fixes these errors?

PR #52, opened 2026-08-31
[cpyrt] Clear stale Python errors before C API calls

aaronj0:
A debug build of CPython asserts when a C API call is made with the error indicator already set. Clear it before the __cpp_cross__ attribute set, the meta_setattro fallthrough to tp_setattro, and the VectorData AddToClass.

From the ROOT migration branch

### DIFF
--- a/src/cpyrt/CPPScope.cxx
+++ b/src/cpyrt/CPPScope.cxx
@@ -274,10 +274,14 @@ static PyObject* pt_new(PyTypeObject* subtype, PyObject* args, PyObject* kwds) {
           // also signals that this is a cross-inheritance class)
           PyObject* bname = cpyrt_PyText_FromString(
               interop::GetBaseName(result->fCppType, 0).c_str());
-          if (PyObject_SetAttrString((PyObject*)result, "__cpp_cross__",
-                                     bname) == -1)
+          if (!bname)
             PyErr_Clear();
-          Py_DECREF(bname);
+          else {
+            if (PyObject_SetAttrString((PyObject*)result, "__cpp_cross__",
+                                       bname) == -1)
+              PyErr_Clear();
+            Py_DECREF(bname);
+          }
         }
       } else if (sz == (Py_ssize_t)-1)
         PyErr_Clear();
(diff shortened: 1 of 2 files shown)


src/cpyrt/Pythonize.cxx
```diff
@@ -1899,6 +1899,7 @@ bool cpyrt::Pythonize(PyObject* pyclass, interop::TCppScope_t scope) {
 
       // data with size
       Utility::AddToClass(pyclass, "__real_data", "data");
+      PyErr_Clear(); // AddToClass might have failed for data
```
guitargeek:
```suggestion
if (!Utility::AddToClass(pyclass, "__real_data", "data"))
    PyErr_Clear();      // no pre-existing 'data' method to alias
```
I think we don't need to do this unconditionally. AddToClass returns a success flag, doesn't it?

src/cpyrt/CPPScope.cxx
```diff
@@ -274,6 +274,7 @@ static PyObject* pt_new(PyTypeObject* subtype, PyObject* args, PyObject* kwds) {
           // also signals that this is a cross-inheritance class)
           PyObject* bname = cpyrt_PyText_FromString(
               interop::GetBaseName(result->fCppType, 0).c_str());
+          PyErr_Clear();
```
guitargeek:
```suggestion
if (!bname) {
    PyErr_Clear();          // string creation failed; skip the annotation
} else {
    if (PyObject_SetAttrString((PyObject*)result, "__cpp_cross__", bname) == -1)
        PyErr_Clear();
    Py_DECREF(bname);
}
```
I would call `PyErr_Clear()` only when an actual happened

src/cpyrt/CPPScope.cxx
```diff
@@ -575,6 +576,7 @@ static int meta_setattro(PyObject* pyclass, PyObject* pyname, PyObject* pyval) {
       meta_getattro(pyclass, pyname); // triggers creation
   }
 
+  PyErr_Clear(); // creation might have failed
```
guitargeek: Also here I would make the `Clear()` conditional on an actual error (I guess it comes from `meta_gettro` here
```c++
if (interop::GetNamed(name, ((CPPScope*)pyclass)->fCppType)) {
    PyObject* attr = meta_getattro(pyclass, pyname);  // triggers creation
    if (!attr) PyErr_Clear();       // creation failed: fall through
    else Py_DECREF(attr);
}
```

guitargeek (changes requested):
Thanks! I'd just suggest to clear the errors only when an error occurred, to not risk clearing other errors that should propagate.

src/cpyrt/Pythonize.cxx
vgvassilev: I believe we need an api that makes an error ignorable (something like llvm::consumeError) where we annotate these functions that can be safe to be ignored. That is, we do something like IgnoreError(cpyrt_PyText_FromString...) where we make sure that we carry the accumulated error before this call and only remove the ones introduced by `cpyrt_PyText_FromString`.

src/cpyrt/CPPScope.cxx
aaronj0: Indeed, nice catch! This was a latent ref leak because we discarded the result of `meta_getattro` without decref

src/cpyrt/Pythonize.cxx
aaronj0: > I think we don't need to do this unconditionally

Yes, thanks for catching this. The original patch did not take this into account..

> I believe we need an api that makes an error ignorable (something like llvm::consumeError)

Yes this would make a lot of sense. Since this will touch all the call sites where we currently clear errors, I will make this a follow up PR with a general helper for this. I have opened an issue tracking this: https://github.com/compiler-research/cppjit/issues/57

src/cpyrt/CPPScope.cxx
aaronj0: Addressed

guitargeek (approved):
LGTM! Thanks for addressing the comments.

PR #54, opened 2026-08-31
[cpyrt] Reuse InitializerListConverter element converters

aaronj0:
SetArg() created an element converter per call and appended it to fConverters, but Clear() frees only fBuffer, so the vector grew without bound across repeated std::initializer_list conversions. Create each element converter once and reuse it.

Upstreamed from the ROOT migration

### DIFF
--- a/src/cpyrt/Converters.cxx
+++ b/src/cpyrt/Converters.cxx
@@ -3186,7 +3186,9 @@ bool cpyrt::InitializerListConverter::SetArg(PyObject* pyobject,
       PyObject* item = PySequence_GetItem(pyobject, i);
       bool convert_ok = false;
       if (item) {
-        Converter* converter = CreateConverter(fValueTypeName);
+        if (i >= fConverters.size())
+          fConverters.emplace_back(CreateConverter(fValueTypeName));
+        Converter* converter = fConverters[i];
         if (!converter) {
           if (CPPInstance_Check(item)) {
             // by convention, use byte copy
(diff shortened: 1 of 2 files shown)


src/cpyrt/Converters.cxx
```diff
@@ -3186,8 +3186,9 @@ bool cpyrt::InitializerListConverter::SetArg(PyObject* pyobject,
       PyObject* item = PySequence_GetItem(pyobject, i);
       bool convert_ok = false;
       if (item) {
-        Converter* converter = CreateConverter(fValueTypeName);
-        if (!converter) {
+        if (fConverters.empty())
+          fConverters.emplace_back(CreateConverter(fValueTypeName));
+        if (!fConverters.back()) {
```
guitargeek:
```suggestion
if (i >= (size_t)fConverters.size())
  fConverters.emplace_back(CreateConverter(fValueTypeName));
Converter* converter = fConverters[i];
if (!converter) {
```
With this change, you'd "absorb" also the `CreateConverter` path below and we don't need to repeat it.

src/cpyrt/Converters.cxx
```diff
@@ -3209,9 +3210,11 @@ bool cpyrt::InitializerListConverter::SetArg(PyObject* pyobject,
             entries += 1;
           }
           if (memloc) {
-            convert_ok = converter->ToMemory(item, memloc);
+            if (i >= fConverters.size()) {
+              fConverters.emplace_back(CreateConverter(fValueTypeName));
+            }
```
guitargeek:
```suggestion
```
Can be dropped with the unified `CreateConverter` path I suggested

guitargeek (changes requested):
Thanks! I think we can unify this to a single "CreateConverter" path though

guitargeek (approved):
LGTM!

PR #56, opened 2026-08-31
[fix] Null-guard the cling.printValue lookup in op_str

conrade-ctc:
`str()` on any instance without an `ostream` inserter segfaults on current
`main`. Commit `5a33027` dropped the faked `cling::runtime::gCling`, and the
pretty-print fallback in `op_str` then dereferenced the failed
`cppjit.gbl.cling` lookup. This guards both lookups, clears the
`AttributeError`, and falls back to the generic repr.

Repro:

```python
import cppjit
cppjit.cppdef("struct Bare { int x; };")
str(cppjit.gbl.Bare())      # SIGSEGV
```

A regression test is included. The crash kills the interpreter, so the test
asserts the fixed behaviour rather than the crash.


src/cpyrt/CPPInstance.cxx
```diff
@@ -875,17 +875,22 @@ static PyObject* op_str(CPPInstance* self) {
     if (!printValue) {
       PyObject* gbl =
           PyDict_GetItemString(PySys_GetObject((char*)"modules"), "cppjit.gbl");
-      PyObject* cl = PyObject_GetAttrString(gbl, (char*)"cling");
-      printValue = PyObject_GetAttrString(cl, (char*)"printValue");
-      Py_DECREF(cl);
+      // no cling namespace exists unless user code declares one
```
aaronj0: Perhaps we can just short-circuit this path completely if we know the interpreter is clang-repl, since that lookup would always return null. We could use the compile-time definition `CPPJIT_USE_CLING`, or a runtime check like in test/support.py

conrade-ctc: Good call — done. The whole pretty-printing block now sits behind `#ifdef CPPJIT_USE_CLING`, so clang-repl builds compile it out and `str()` falls straight through to the generic repr; `strings` on the clang-repl `libcppjit.so` shows the `printValue` lookup gone, and the object shrinks slightly. One thing the gate needed: `CPPJIT_USE_CLING` was only ever a CMake option and never reached the sources, so I added the matching `target_compile_definitions` entry — without it the `#ifdef` would have been false on the cling build too. The null-guards stay inside the cling branch, in case an odd cling state leaves the lookup empty. Worth naming the trade-off: a user-declared `namespace cling { printValue }` is no longer honored on clang-repl, which I think is right given the native value-printing direction in compiler-research/CppInterOp#1100.

src/cpyrt/CPPInstance.cxx
```diff
@@ -869,23 +869,30 @@ static PyObject* op_str(CPPInstance* self) {
   }
 
   // 2. Cling's pretty printing (not done through backend for performance
-  // reasons)
+  // reasons). Cling only: clang-repl has no cling namespace to look up, so the
+  // whole path compiles out and str() falls through to the generic repr.
+#ifdef CPPJIT_USE_CLING
```
vgvassilev: This should probably be required to CppInterOp’s toString interface which will still crash for clang-repl but this time we can fix it there.

conrade-ctc: Happy to make that change here: route str() through Cpp::ObjToString and drop the cling-only block. clang-repl would then keep failing in toString until compiler-research/CppInterOp#1100 lands the fix there. Do you want that in this PR, or should we keep the short-circuit and switch once #1100 is fixed?

src/cpyrt/CPPInstance.cxx
```diff
@@ -868,23 +868,30 @@ static PyObject* op_str(CPPInstance* self) {
...
```
vgvassilev: I believe if we want to land this ifdef-based infrastructure we probably want to put it in a single compatibility-like file that expands the right ifdefs in one place.

conrade-ctc: I went ahead and moved this into a compatibility header

vgvassilev (dismissed):
Lgtm!

CMakeLists.txt
```diff
@@ -76,6 +76,8 @@ target_compile_definitions(cppjit PRIVATE
     CPPINTEROP_INCLUDE_DIR="${CPPJIT_INTEROP_RUNTIME_INCLUDES}"
     CPPJIT_CLANG_MAJOR="${CPPJIT_INTEROP_CLANG_MAJOR}"
     CPPJIT_CLANG_INCLUDE_DIR="${CPPJIT_INTEROP_CLANG_DIR}"
+    # cling-only code paths need the flavor at compile time, not just in cmake
```
aaronj0: We can probably drop this comment here.

test/test_regression.py
```diff
@@ -1648,3 +1648,58 @@ def test52_no_cpp_name_for_template_arg(self):
 
         with raises(TypeError):
             cppjit.gbl.std.vector[object()]
+
+    def test53_str_fallback_without_ostream_insertion(self):
```
aaronj0, line 1648: We do have existing tests that are marked crashing with clang-repl due to ToString. I am not sure if adding a Python test that spawns a process guaranteed to crash is a good idea, especially since this does not live on the API surface, and this test suite is meant to be user-facing. Imo this can be moved into a CppInterOp only reproducer. If `compat::GetClingPrintValue()` returns null, can we prevent the crash and gracefully error out instead?

aaronj0 (dismissed):
LGTM % the comments

CMakeLists.txt
conrade-ctc: Dropped.

test/test_regression.py
conrade-ctc: The null return was already safe. `op_str` falls through to the generic repr there.

The crash the test caught happens one step earlier. `op_str` calls `Cpp::ObjToString`, and `Interpreter::toString` is an `assert(0)` stub. It aborts with assertions on and returns `""` without them. The stub is not backend specific, so the existing markers say "and on Cling".

I moved that call behind the same compatibility header and made it return the empty string on every build, so `op_str` reaches the generic repr everywhere. #1100 turns the helper back into a one-line forward.

The subprocess test is gone. The new `test53` runs in process and checks that `str(obj)` equals `repr(obj)`. The three xfail markers that named this crash now pass on an assertions-on build, so I removed them.

aaronj0 (approved):
LGTM!

test/test_regression.py
aaronj0: Nice, Thanks!

PR #58, opened 2026-09-01
Introducing proper deallocation ways for CPPInstance

keremsahn:
Built upon #37, just last commit is related to this PR
I added three flags to the CPPInstance to distinguish different allocation methods, and in HandleReturn ,where ownership is given, corresponding flags are set after analyzer/attribute check.

Using these new flags CPPInstance::op_dealloc_nofree introduces different kind of deallocations.
Also tests changed a lot after this feature

And I need to explain why I added kIsArrayAlloc instead of using kIsArray, users can set kIsArray flag by reshape feature:
```cpp
Klass* alloc(size_t count){  return ::operator new(sizeof(Klass)*count);}
```
and if user reshapes the returned object, that will lead UB because object is deallocated by operator delete[]

@Vipul-Cariappa @aaronj0 @vgvassilev

### DIFF
--- a/src/cpyrt/CPPInstance.h
+++ b/src/cpyrt/CPPInstance.h
@@ -27,21 +27,25 @@ typedef std::vector<std::pair<ptrdiff_t, PyObject*>> CI_DatamemberCache_t;
 class CPPInstance {
 public:
   enum EFlags {
-    kDefault = 0x0000,
-    kNoWrapConv = 0x0001,  // use type as-is (eg. no smart ptr wrap)
-    kIsOwner = 0x0002,     // Python instance owns C++ object/memory
-    kIsExtended = 0x0004,  // has extended data
-    kIsValue = 0x0008,     // was created from a by-value return
-    kIsReference = 0x0010, // represents one indirection
-    kIsArray = 0x0020,     // represents an array of objects
-    kIsSmartPtr = 0x0040,  // is or embeds a smart pointer
-    kIsPtrPtr = 0x0080,    // represents two indirections
-    kIsRValue = 0x0100,    // can be used as an r-value
-    kIsLValue = 0x0200,    // can be used as an l-value
-    kNoMemReg = 0x0400,    // do not register with memory regulator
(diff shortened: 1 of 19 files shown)


keremsahn:
I do not have access to a mac and I am trying to figure out why macos tests fail, that is why I am pushing constantly

test/test_memoryanalysis.py
```diff
@@ -101,6 +126,134 @@ def test08_analyzer_off_but_cache(self):
...
+        cppjit.gbl.memory.memOwn.dtorCount = 0
+
+        def alloc():
+            obj = cppjit.gbl.memory.allocOperatorNewArrAttr(5)
+            assert obj.__python_owns__
+            assert obj.__is_no_construct__
+            assert obj.__is_array_alloc__
+
+        alloc()
+        assert cppjit.gbl.memory.memOwn.dtorCount == 0
```
Vipul-Cariappa, line 208:
```suggestion
        alloc()
        gc.collect()
        assert cppjit.gbl.memory.memOwn.dtorCount == 0
```

aaronj0:

src/cpyrt/CPPInstance.h
```diff
@@ -27,21 +27,25 @@ typedef std::vector<std::pair<ptrdiff_t, PyObject*>> CI_DatamemberCache_t;
...
+    kIsReference = 0x00010,  // represents one indirection
+    kIsArray = 0x00020,      // represents an array of objects
+    kIsSmartPtr = 0x00040,   // is or embeds a smart pointer
+    kIsPtrPtr = 0x00080,     // represents two indirections
+    kIsRValue = 0x00100,     // can be used as an r-value
+    kIsLValue = 0x00200,     // can be used as an l-value
+    kNoMemReg = 0x00400,     // do not register with memory regulator
+    kIsRegulated = 0x00800,  // is registered with memory regulator
+    kIsActual = 0x01000,     // has been downcasted to actual type
+    kHasLifeLine = 0x02000,  // has a life line set
```
Vipul-Cariappa: Can we restore these changes?

PR #59, opened 2026-09-01
[ci] Harden the wheels workflow and cibuildwheel configuration

aaronj0:
Various suite of mechanical and workflow improvements based on some research on the standards followed by scipy/numpy/pybind11:

Workflow hardening:
- `persist-credentials: false` on every checkout
- `if-no-files-found: error` on every artifact upload
- `zizmor.yml` records the action-pin policy

Artifact integrity:
- Every wheel is twine-checked via cibuildwheel's audit hook and validated against a zip-member allowlist on all platforms, so nothing unexpected ships.
- The build must leave the checkout unmutated (`git diff --exit-code`, plus the pybind-template check).
- The sdist is twine-checked.

sdist is now tested:
- New `test-sdist` job stages the recipe LLVM, pip-installs the tarball, and runs the smoke and full suites. This matches what numpy and scipy do.

cibuildwheel config fixes:
- `PYTHONSAFEPATH=1` in the test env, so imports resolve from the installed wheel rather than the source tree.
- Host-mount disabled in the build container.
- macOS requirements install moved to `before-test`.
- sdist excludes added.
- `cp314t`/`cp315` exclusions documented as policy (verified against v4.2.0 defaults).
- `archs = ["auto64"]` with the macos-x86_64 cell: native runners only, no cross-compilation.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,7 @@ minimum-version = "build-system.requires"
 wheel.install-dir = "."
 wheel.packages = ["python/cppjit"]
 cmake.build-type = "Release"
+sdist.exclude = [".github", ".gitignore", ".clang-format"]
 
 [[tool.dynamic-metadata]]
 provider = "scikit_build_core.metadata.regex"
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -25,36 +25,45 @@ minimum-version = "build-system.requires"
...
 
 [[tool.dynamic-metadata]]
 provider = "scikit_build_core.metadata.regex"
 field = "version"
 input = "python/cppjit/_version.py"
 
 [tool.cibuildwheel]
+# The build list is policy, not a glob accident: cp314t waits on a
+# free-threading audit (Py_mod_gil, import must not re-enable the GIL);
+# cp315 joins at the 3.15 release, not at the rc cibuildwheel builds.
```
guitargeek: I thought the Python developers recommend the package maintainers to build wheels as soon as the Python release candidate is out? Python 3.15 RC 1 is out already I think.

guitargeek (approved):
Thanks a lot!

Consider adding Python 3.15 wheels as well (in a follow-up PR) since the release candidate was published last month: https://blog.python.org/2026/08/python-3150-rc1/

pyproject.toml
aaronj0: Indeed a good idea! I'll follow up with a 3.15 rc wheel matrix. That will need extending the ci-workflows wheels job as we maintain the build-images on that repo. Do you want to drop the policy comments?

aaronj0: Comments have been cleaned up

PR #61, opened 2026-09-01
[test] Skip the dictionary tests without a host toolchain

conrade-ctc:
On a host with no make and no C++ compiler, every module that builds a test dictionary errors out in setup_make:

    FileNotFoundError: [Errno 2] No such file or directory: 'make'

This is a follow-up to #51. setup_make now checks for make and for make's $(CXX) first, and skips the module with a clear reason. A build system that sets CPPJIT_TEST_SKIP_MAKE is unaffected. In test_basic_api only test03_add_library_path loads a dictionary, so it gets a skip marker and the other API tests keep running.

### DIFF
--- a/test/support.py
+++ b/test/support.py
@@ -1,10 +1,12 @@
 from __future__ import print_function
 
 import os
+import shutil
 import subprocess
 import sys
 
 import py
+import pytest
 
 try:
     import fcntl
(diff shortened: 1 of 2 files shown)


aaronj0 (dismissed):
LGTM, thanks!

PR #62, opened 2026-09-01
[interop] Pass template names and named constants as template arguments

conrade-ctc:
A template argument can now name a template or a constant. A class or alias template name becomes a template-template argument. A constexpr variable or an enum constant becomes an expression, so Sema reads its value. `true` and `false` become bool values. The first commit fixes a separate bug: `is_integral` rejected a leading minus, so a negative literal became a type name.

The by-name paths need CppInterOp compiler-research/CppInterOp#1074, which resolves the name. Before this merges, the CppInterOp pin moves to the commit that carries #1074 and the capability probe goes away. Do not merge this before #1074.


src/interop/interop_wrapper.cxx
```diff
@@ -469,6 +471,22 @@ static bool is_identifier(std::string_view s) {
...
 
+// A template argument carried by name needs a CppInterOp that resolves the
+// name; such a CppInterOp exports SupportsNamedTemplateArguments.
+static bool supportsNamedTemplateArgs() {
+#ifdef _WIN32
+  return false; // no dlsym; enable once the pin guarantees the capability
+#else
+  static const bool Supported = [] {
+    // CppInterOp is dlopen'ed RTLD_LOCAL; its exports need its own handle.
+    void* handle = dlopen(cppinterop_paths().Library.c_str(),
```
vgvassilev, line 482: This seems a weird way to check if a feature is there or not. Can’t we rely on the version hash, for example?

conrade-ctc: The probe guards a window, not a permanent difference. cppjit pins CppInterOp by commit sha. The current pin (9802d619) predates #1074, so `Cpp::GetNamed` does not resolve a named constant yet. Without the guard, the by-name tests fail on the pinned build.

I looked at the version hash. `Cpp::GetVersion()` reads the VERSION file, which is the same before and after #1074. `Cpp::GetBuildInfo()` has no git sha, and a sha has no order, so a check cannot express "at least this commit".

So I think the pin is already the version check. My proposal: drop the probe and hold this PR until #1074 merges. Then I rebase, move the pin to a commit that carries #1074, and the by-name tests run unconditionally. This matches #63, which waited on #1101 the same way. The removal is ready locally.

Tell me if you prefer a different order.

PR #63, opened 2026-09-01
[python] Report the dlopen failure reason from load_library

conrade-ctc:
Cpp::LoadLibrary drops the loader's failure reason, so load_library raises a bare error. compiler-research/CppInterOp#1101 adds an optional reason out-parameter to Cpp::LoadLibrary. This change passes a std::string to it and reports the text in the RuntimeError. The CppInterOp pin moves to the commit that carries #1101. It adds a regression test for a missing library and a truncated ELF header.


aaronj0 (commented):
This is a nice improvement! Perhaps we could have Cpp::LoadLibrary give us the diagnostics on why it failed so we can avoid re-attempting dlopen with ctypes on the Python side.

edit: Just saw https://github.com/compiler-research/CppInterOp/pull/1101, I think that would be the best solution here

conrade-ctc:
Agreed, #1101 is the right place for this, and I think a `LoadLibrary` that hands back the reason is the proper end state. #63 already prefers the captured stderr text and only falls back to the ctypes probe when that text is empty, so with #1101 in the pin the probe is idle. I kept it because the pin lags CppInterOp and the exception is what a notebook user sees, not stderr. Two options, your call: merge #63 as the interim and I remove the probe when the pin catches up, or I close it and follow #1101 with a CppInterOp API that returns the reason (an out-parameter overload of `Cpp::LoadLibrary`, say) and wire cppjit to that. I am happy either way.

aaronj0:
> Agreed, #1101 is the right place for this, and I think a `LoadLibrary` that hands back the reason is the proper end state. #63 already prefers the captured stderr text and only falls back to the ctypes probe when that text is empty, so with #1101 in the pin the probe is idle. I kept it because the pin lags CppInterOp and the exception is what a notebook user sees, not stderr. Two options, your call: merge #63 as the interim and I remove the probe when the pin catches up, or I close it and follow #1101 with a CppInterOp API that returns the reason (an out-parameter overload of `Cpp::LoadLibrary`, say) and wire cppjit to that. I am happy either way.

We can add an optional out param to LoadLibrary and once that lands, use that in this PR (you can bump the pinned commit here so it builds with latest CppInterOp containing #1101)

conrade-ctc:
Opened compiler-research/CppInterOp#1107 with the optional `std::string* error` out-parameter, stacked on #1101. The cppjit side is pushed here already: `load_library` passes a `std.string` and reports its text, and the ctypes probe is gone. CI on this PR stays red until the pin can point at a CppInterOp commit that carries #1107; I will bump `CPPINTEROP_GIT_TAG` and squash as soon as it lands.

test/test_basic_api.py
```diff
@@ -59,6 +60,24 @@ def test03_add_library_path(self):
             shutil.copyfile(test_dct + ".so", tpath + "/test.so")
             cppjit.load_library("test.so")
 
+    @mark.skipif(IS_LINUX == 0, reason="checks Linux dlerror text")
```
aaronj0: can we make this `xfail` and `not IS_LINUX`

conrade-ctc: Done: the test now runs everywhere and is an `xfail` off Linux.

conrade-ctc: Turns out macOS passes too (strict xfail flagged it as XPASS), so I dropped the marker; the test now runs everywhere.

PR #64, opened 2026-09-01
[cpyrt] Expose method const-qualification in func_overloads_types

conrade-ctc:
A caller could not tell whether a given overload was const-qualified. Each overload's entry in `func_overloads_types` now carries an `is_const` field. Free functions and static methods report `is_const` as `False`.


aaronj0 (approved):
LGTM!

PR #66, opened 2026-09-02
[interop] Filter deleted functions from Python-visible overload sets

aaronj0:
A deleted function is not callable, yet it still entered the overload set and contributed a spurious conversion error to every failed-call report. That defeats SetDetailedException's rule that failures which are all C++ exceptions of one type are re-raised as that type.

Fixes tests in the ROOT migration as well as `test34_no_ctors_in_base` which expects TypeError for a class whose only constructors are deleted.

### DIFF
--- a/src/interop/interop_wrapper.cxx
+++ b/src/interop/interop_wrapper.cxx
@@ -1173,16 +1173,31 @@ ptrdiff_t interop::GetBaseOffset(TCppScope_t derived, TCppScope_t base,
 }
 
 // method/function reflection information ------------------------------------
+// A deleted overload is not callable, and leaving it in the set adds a
+// spurious conversion error to every failed-call report.
+static void
+remove_deleted_methods(std::vector<interop::TCppMethod_t>& methods) {
+  methods.erase(std::remove_if(methods.begin(), methods.end(),
+                               [](interop::TCppMethod_t m) {
+                                 return Cpp::IsFunctionDeleted(m);
+                               }),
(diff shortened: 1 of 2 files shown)


src/interop/interop_wrapper.cxx
```diff
@@ -1173,16 +1173,31 @@ ptrdiff_t interop::GetBaseOffset(TCppScope_t derived, TCppScope_t base,
 }
 
 // method/function reflection information ------------------------------------
+// A deleted overload is not callable, and leaving it in the set adds a
+// spurious conversion error to every failed-call report.
+static void
+remove_deleted_methods(std::vector<interop::TCppMethod_t>& methods) {
+  methods.erase(std::remove_if(methods.begin(), methods.end(),
+                               [](interop::TCppMethod_t m) {
+                                 return Cpp::IsFunctionDeleted(m);
```
vgvassilev, line 1182: I guess here the question for @Vipul-Cariappa is if the new system suggests `deleted` functions and is that what general c++ do?

Vipul-Cariappa: I believe this is filtering the lookup and not the overload resolution. This should be fine.

vgvassilev (approved):
LGTM!

src/interop/interop_wrapper.cxx
vgvassilev: Ok, thank you!

PR #67, opened 2026-09-02
[cpyrt] Raise TypeError when a template argument names no C++ type

aaronj0:
Every Python object has a __name__, so an object naming no C++ type reached AddTypeName's fallback, where GetType() returned null; that null was then pushed as a template argument and dereferenced by IsEnumType. Skip such an argument, and report the string-based path's failure as TypeError like the type-based one already does.


guitargeek (approved):
LGTM!

PR #70, opened 2026-09-02
[cpyrt] Do not decref a constructor result on Windows

aaronj0:
ExecuteFast() ends with a Windows-only block that drops the result when a Python exception is pending. It is shared by every CPPMethod subclass, including CPPConstructor, whose executor returns the address of the new C++ object cast to PyObject*; decref'ing that corrupts the heap. Ask the method whether its executor hands back a real PyObject* first. The object leaks instead, on the error path of a call that is already failing.


guitargeek (approved):
LGTM! The undefined behavior fixes are indeed not controversial.

PR #71, opened 2026-09-02
[cpyrt] Restore support for non-const void*& arguments

aaronj0:
VoidPtrRefConverter was registered only for "const void*&", so a mutable void*& fell under the general ban on non-const pointer references. Passing an object through an opaque void*& handle is supported by design; the exact-match factory lookup runs before the T*& rejection, which stays in effect for typed pointer references.

From the ROOT migration


guitargeek (approved):
LGTM! Thank you.

PR #74, opened 2026-09-03
[cpyrt] Add CStringArrayConverter::ToMemory

aaronj0:
Assigning to a const char** data member inherited the buffer-only ToMemory of the base converter, so a Python string was rejected with a property type mismatch. Accept a string as well, with the buffer copy factored into a helper covering the fixed and unbounded cases and the lifetime tie-in. From the ROOT migration.


guitargeek (approved):
LGTM, thanks!

PR #75, opened 2026-09-03
[cpyrt] Detect enum arguments by type in overload priority

aaronj0:
GetPriority() deprioritizes enum arguments so that a competing integer overload wins, since C++ has no implicit int->enum conversion, but it detected them with IsEnumScope(GetScope(name)). GetScope does not resolve an enum name to a scope, so the penalty never applied. Query the argument type with IsEnumType as well, per the existing FIXME.

### DIFF
--- a/src/cpyrt/CPPMethod.cxx
+++ b/src/cpyrt/CPPMethod.cxx
@@ -581,7 +581,9 @@ int cpyrt::CPPMethod::GetPriority() {
       if (scope)
         priority += static_cast<int>(interop::GetNumBasesLongestBranch(scope));
 
-      if (interop::IsEnumScope(scope))
+      // GetScope does not resolve enum names, so also detect enums by type
+      if (interop::IsEnumScope(scope) ||
+          interop::IsEnumType(interop::GetMethodArgType(fMethod, iarg)))
         priority -= 100;
 
       // a couple of special cases as explained above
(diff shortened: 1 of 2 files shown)


guitargeek (changes requested):
Thanks, this goes in the right direction! `pick(2)` now returns 2, penalty fires via `IsEnumType(GetMethodArgType(...))`. One remaining problem though: the penalty is argument-agnostic, so `pick(Color.Green)` now returns 2: previously 1, and C++ overload resolution would also give 1. An enum object is an exact match for the enum overload.

I think this is a behavioral regression that needs to be fixed and locked in to the unit test.

aaronj0:
> Thanks, this goes in the right direction! `pick(2)` now returns 2, penalty fires via `IsEnumType(GetMethodArgType(...))`. One remaining problem though: the penalty is argument-agnostic, so `pick(Color.Green)` now returns 2: previously 1, and C++ overload resolution would also give 1. An enum object is an exact match for the enum overload.
> 
> I think this is a behavioral regression that needs to be fixed and locked in to the unit test.

Makes sense. Since the priority alone can't tell the two calls apart, I've addressed this distinction by moving it into the converter logic itself.  By-value enum parameters now get an `EnumConverter` that takes an instance of the same enum as an exact match in the first round and delegates the actual conversion to the underlying integer converter, while the integer converters defer enum instances to the implicit round, as they already do for bools. So now in our test example, `pick(Color.Green)` reaches the enum overload before the integer one is allowed to accept it, `pick(2)` still goes to `pick(unsigned int)`, and an enum passed to a plain integer parameter still converts when there is no enum overload.

PR #78, opened 2026-09-03
[ci] Cover all supported Pythons and both LLVM majors per PR

aaronj0:


mcbarton:
I see xpassed in the test results comment above. I thought this https://github.com/compiler-research/cppjit/blob/8be966d32d6f9e6dcddd296f86dfdbb7edbb86d2/pyproject.toml#L64 was supposed to make the ci fail if that happens? I don't know if it points towards what the issue is, but the xpasses are not tripping the ci are the macos x86 jobs.

aaronj0:
Behaves as expected. As documented on that test: flaky on intel macs as it checks for a delta RSS threshold which is not deterministic and can fail intermittently. The xfail_strict = true you've pointed to in pyproject.toml works correctly, test08_list_creation is the only test that overrides it conditionally on intel mac. (see the marker). So if any test on any other platform xpasses, it would in fact, trip the CI

PR #79, opened 2026-09-04
[build] Consume an external CppInterOp via find_package

aaronj0:
Supersedes #48, makes the build system robust for purely developement builds:

#### Development build (CMake)

The build tree runs in place; nothing is installed. CppInterOp builds
and stages inside it, and the python package assembles under
`<build>/python`:

```bash
cmake -S cppjit -B build -DLLVM_DIR=$LLVM_DIR
cmake --build build -j
export PYTHONPATH=$PWD/build/python
```

A prebuilt CppInterOp, either an install prefix or a build directory,
is consumed in place through `CppInterOp_DIR` instead of being rebuilt:

```bash
cmake -S cppjit -B build -DLLVM_DIR=$LLVM_DIR \
    -DCppInterOp_DIR=$PWD/CppInterOp/build/lib/cmake/CppInterOp
```

#### Development build (pip editable)

Editable install with a persistent build directory; a one-file change
rebuilds incrementally:

```bash
pip install --no-build-isolation -ve . \
    --config-settings=build-dir=build \
    --config-settings=cmake.define.LLVM_DIR=$LLVM_DIR
```

To co-develop CppInterOp alongside cppjit, point the build at a local
checkout; it overrides the pinned tag:

```bash
git clone https://github.com/compiler-research/CppInterOp.git ../CppInterOp
pip install --no-build-isolation -ve . \
    --config-settings=build-dir=build-local \
    --config-settings=cmake.define.LLVM_DIR=$LLVM_DIR \
    --config-settings=cmake.define.CPPINTEROP_SOURCE_DIR=$PWD/../CppInterOp
```

Keep the checkout API-compatible with the pinned `CPPINTEROP_GIT_TAG` in
`CMakeLists.txt`.

#### Verify the install

Run from a directory outside the checkout; the in-tree `python/cppjit`
shadows the installed extension:

```bash
cd /tmp && python -c "import cppjit
cppjit.cppdef('int f(int x) { return x + 1; }')
print(cppjit.gbl.f(41))"   # 42
```

#### Tests

```bash
pip install -r requirements.txt
sudo apt-get install -y libboost-dev libeigen3-dev  # optional; those tests skip without them
cd test
make -j4                          # builds the *Dict.so dictionaries the tests load
python -m pytest -ra --tb=short
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,6 @@ set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
-include(ExternalProject)
 include(GNUInstallDirs)
 
 # This option won't make a lot of sense since we only ship the shared library in site-packages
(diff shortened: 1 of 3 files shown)


guitargeek (approved):
LGTM! That works for me too!

PR #81, opened 2026-09-04
[build] Add PyPI project metadata: readme, urls, classifiers

aaronj0:


pyproject.toml
```diff
@@ -6,8 +6,23 @@ build-backend = "scikit_build_core.build"
 name = "cppjit"
 dynamic = ["version"]
 description = "CppJIT: fast and automatic Python-C++ interoperability"
+readme = "README.md"
 license = "BSD-3-Clause-LBNL"
 requires-python = ">=3.12"
+classifiers = [
+    "Development Status :: 3 - Alpha",
+    "Intended Audience :: Developers",
+    "Intended Audience :: Science/Research",
```
vgvassilev, line 15: How about General Use?

aaronj0: The only other generic options I could find are "Other Audience" and "End Users/Desktop", I am not sure if that is meant for our library. Maybe "Other Audience" but End Users/Desktop does not apply in our case.
Developers + Science/Research seems to be the category used by developer tools

vgvassilev: Adding developers makes sense to me.

aaronj0: Addressed

vgvassilev (approved):
Lgtm!

PR #82, opened 2026-09-04
[build] Make the build tree runnable without an install step

aaronj0:
Follow up to #79 that enables the standard development build, without relying on the install layout. This patch is a simple refactor of our current cmake when driven by packaging (SKBUILD) and standard cmake


guitargeek (approved):
LGTM!

PR #84, opened 2026-09-06
Add PyPI publishing job and release process

aaronj0:


docs/ReleaseNotes.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Contributors
+
+Special thanks to everyone who contributed to this release:
+
+- ...
+
+<!-- Contributors since the previous release, with commit counts, in
+alphabetical order:
+ git log --pretty='%an' <previous-tag>..HEAD | sort | uniq -c
```
aaronj0, line 42: <previous-tag> here will be replaced with the last tag after the first release

docs/ReleaseProcess.md
```diff
@@ -0,0 +1,75 @@
...
+   through the manual fallback below.
+
+4. Create the GitHub release for the tag. The body is the finalized
+   `docs/ReleaseNotes.md`. The attached artifacts are the uploaded
+   wheels and sdist.
+
+5. Open the post-release PR. It sets `_version.py` to the next
+   version with a `.dev0` suffix (e.g., `0.1.0a2.dev0` follows
+   `0.1.0a1`) and resets `docs/ReleaseNotes.md`.
+
```
vgvassilev:
```suggestion
5. Announce the release on the compiler-research website and  
    the relevant mailing lists.

6. Open the post-release PR. It sets `_version.py` to the next
   version with a `.dev0` suffix (e.g., `0.1.0a2.dev0` follows
   `0.1.0a1`) and resets `docs/ReleaseNotes.md`.

```

docs/ReleaseProcess.md
```diff
@@ -0,0 +1,75 @@
+# Release Process
+
+A cppjit release is a git tag on main. The tag push builds the wheels,
+and the PyPI workflow uploads what that build produced. Future releases
+will extend this to more distribution channels like conda.
+
+## Procedure
+
+1. Open the release PR:
+   - Set `__version__` in `python/cppjit/_version.py` to the release
```
vgvassilev, line 10: How does cmake know what is the current version? Does it need to parse that? We should probably have some notion of a simple `VERSION` file.

pyproject.toml
```diff
@@ -5,7 +5,7 @@ build-backend = "scikit_build_core.build"
 [project]
 name = "cppjit"
 dynamic = ["version"]
-description = "CppJIT: fast and automatic Python-C++ interoperability"
+description = "Automatic Python-C++ interoperability layer and bindings generator"
```
vgvassilev, line 8: Consider:

```suggestion
description = "An automatic Python–C++ interoperability layer and bindings generator built on LLVM's Clang-REPL"
```

Consider also adding '"exascale", "scientific-computing",' to the keywords.

Consider adding the following topics to the github repo: python, cpp, llvm, clang, clang-repl, jit, jit-compilation, python-bindings, bindings, interoperability, ffi, cppyy, cppinterop

vgvassilev (dismissed):
LGTM % comments

docs/ReleaseProcess.md
aaronj0: The single source is python/cppjit/_version.py, which the build system reads with the regex declared in pyproject.toml when building the wheel/sdist, and provides `cppjit.__version__` at runtime. This is the recommended practice for Python packages: https://packaging.python.org/en/latest/discussions/single-source-version/

cppyy also followed the same practice, setup.py regex-reads python/cppyy/_version.py and cites the same PyPA guide: https://github.com/wlav/cppyy/blob/2d28e09af72dd7f951c341d74266a0c31599d8a8/setup.py#L36

 If we need the info in CMake (a build stamp etc.) we can have it read the same file with a one-line string(REGEX MATCH), so I don't think we need a separate VERSION file, that would just be a second copy to keep in sync.

vgvassilev (approved):
LGTM!

PR #87, opened 2026-09-07
[cmake] Keep libclangCppInterOp symbols in debug-info builds

aaronj0:
Apply a check on the build type so that we are not always building the stripped CppInterOp target in debuginfo builds.


vgvassilev (approved):
Lgtm!

PR #88, opened 2026-09-07
Release  v0.1.0a1

aaronj0:


docs/ReleaseNotes.md
```diff
@@ -1,43 +1,59 @@
...
+- Compiler-derived bindings: string-based type lookups give way to
+  opaque handles on the Clang AST, and type conversion and overload
+  resolution follow Clang's own rules.
+- One package: the cppyy, CPyCppyy, and cppyy-backend distribution
+  stack is consolidated into a single package and repository, built
+  with scikit-build-core and published to PyPI through trusted
+  publishing.
+- Supported platforms: CPython 3.12-3.14 on Linux x86_64/aarch64 and
+  macOS arm64/x86_64. cppyy's PyPy and Windows support is not carried
+  over in this release.
```
guitargeek: Does that mean you plan to add Windows support later, or is is dropped for good? I think it's not clear to the reader

vgvassilev: Maybe we can open and issue and say window support is tracked here and reference the issue.

aaronj0: Ah, yes. That section is meant to only refers to how this release differs from cppyy. I can add "Windows is not supported with this release but is a planned future development" if that makes it more clear

mcbarton: Just a comment on Windows support. If someone can fix the emulated tls symbol issue in llvm, then I should (hopefully) be able to get cppjit buildable on Windows, given the result of my experiments trying to get cppyy built on Windows.

vgvassilev: Another good reason to track this in a separate issue :)

guitargeek (approved):
LGTM! Thanks for mentioning that Windows support is planned now.

PR #90, opened 2026-09-07
[cpyrt] Use GetActualClass in cast_actual

aaronj0:
cast_actual kept the static class, so comparing two proxies of the same object through different base subobjects compared unequal addresses. Reachable when a proxy is not downcast on creation, e.g. for a pinned type. Also drops a stray '&' from the ostream operand name used for the operator<< lookup, and fixes two comment typos.

From the ROOT migration.


guitargeek (approved):
LGTM! Thanks for porting this over.

PR #91, opened 2026-09-07
[ci] Improve nightlies by dropping trigger workflow, fix naming-scheme

aaronj0:
Clean up CI naming scheme and drop redundant pr-nightly trigger. The mechanism works quite well when tested on open PR's (adding `build-wheels` or `test-nightly` instantly triggers extended runs for more coverage)


.github/workflows/ci.yml
```diff
@@ -1,4 +1,4 @@
-name: CI
+name: Build & Test
```
vgvassilev: Doesn’t that make the ci status row much longer?

aaronj0: It should be a net shorten: 
`CI / macos-26-intel/llvm21/py3.14/c++20 / Build & Test`
becomes
`Build & Test / macos-26-intel/llvm21/py3.14/c++20`

vgvassilev: Ah, ok, isn't the CI implying "Build & Test"?

aaronj0: Yeah.. there is some redundancy there as it now shows it twice. The issue is the parent workflow in ciwf is named "Build & Test". Probably needs some change there. Do you want to drop "Build & Test"? I don't have any preference and would ideally like the smallest name with the relevant info without much verbosity/length

aaronj0: I guess CI is a global term that is technically correct for any workflow, hence that's wouldn't differentiate testing from wheels and nightlies. We can probably have `Test|Wheels|Nightlies` as the category in front and the `platform/arch/llvm/python` as a consistent pattern that follows. How is that?

We would then have:

```
Test / macos-26-intel/llvm21/py3.14/c++20
Wheels / ubuntu-24.04/llvm22/py3.14/c++20
Nightly / ubuntu-24.04-arm/llvm21/py3.14/c++20
```

which to me is minimal in length with complete info

vgvassilev: I am fine as long as it is the shortest meaningful set which allows me to view these things on a reasonably-sized smart phone.

PR #94, opened 2026-09-08
[cpyrt] Cache anonymous-enum constant values in CPPDataMember

aaronj0:
dm_get computed the value of an anonymous-enum constant but did not cache it. kIsEnumPrep is cleared on the first access and kIsEnumType was never set, so later accesses skipped the enum branch and read the instance memory at the member's offset instead. Cache the value in fDescription and set kIsEnumType, as the named-enum path does. From the ROOT migration.


guitargeek (approved):
LGTM!

PR #95, opened 2026-09-08
[build] Add CPPJIT_EXTRA_INTERPRETER_ARGS for packagers

guitargeek:
The embedded clang probes well-known prefixes for the host toolchain's headers. Where a distribution does not lay its toolchain out that way -- Nix, Guix, any relocatable prefix -- the probe comes up empty and the interpreter fails at its first #include, with no recourse short of patching the wrapper. Bake the search paths in at configure time instead. CppInterOp still appends CPPINTEROP_EXTRA_INTERPRETER_ARGS after them, so the runtime can add to the baked arguments.

The value is escaped for the string literal it lands in, so a Windows include path survives the preprocessor, and a ';' is rejected as CMake's list separator. Report a failed interpreter creation while here: a rejected argument left CreateInterpreter returning nullptr while LoadCppInterOp reported success, and the first code to assume an interpreter segfaulted.


aaronj0:
I am not sure if deriving the interpreter args from a build time variable is a good idea when it is a run time operation. Then, the interpreter standard becomes a part of how you package cppjit, when it is something that can be dynamically configured.

guitargeek:
Indeed, the place for a patch like this should rather be CppInterOp.

PR #97, opened 2026-09-08
[cpyrt] Import the cppjit module when gThisModule is null

aaronj0:
Entering through the public API (Instance_FromVoidPtr) without importing cppjit leaves gThisModule null. With Python already initialized, Initialize() skipped the import and reported success anyway, so CreateScopeProxy() dereferenced null. Import the extension module whenever gThisModule is null.

### DIFF
--- a/src/cpyrt/API.cxx
+++ b/src/cpyrt/API.cxx
@@ -59,9 +59,15 @@ static bool Initialize() {
                 << std::endl;
       return false;
     }
+  }
 
-    // force loading of the cppjit module
-    PyRun_SimpleString(const_cast<char*>("import cppjit"));
+  // Importing the extension module is what runs the cpyrt initialization
+  // that sets gThisModule.
+  if (!cpyrt::gThisModule) {
+    PyObject* cppjitmod = PyImport_ImportModule("cppjit");
+    if (!cppjitmod)
+      return false;
+    Py_DECREF(cppjitmod);
   }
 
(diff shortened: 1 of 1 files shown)


src/cpyrt/API.cxx
```diff
@@ -59,9 +59,15 @@ static bool Initialize() {
                 << std::endl;
       return false;
     }
+  }
 
-    // force loading of the cppjit module
-    PyRun_SimpleString(const_cast<char*>("import cppjit"));
+  // Importing the extension module is what runs the cpyrt initialization
+  // that sets gThisModule.
+  if (!cpyrt::gThisModule) {
```
vgvassilev, line 66: Can we open an issue tracking the progress of removing global state?

guitargeek (approved):
LGTM!

src/cpyrt/API.cxx
guitargeek: Sure, please go ahead! Would be useful if the issue said which globals you have in mind (`gThisModule`, `gPyTypeMap`, `gMainDict`, the `PyString` globals? ..) and what problem they cause (subinterpreters, teardown, etc.), so it's actionable rather than a general cleanup item. Because in general, global "immortal" state is not an anti-pattern in CPython.

aaronj0: I'm currently investigating this and will open an issue with more details

PR #100, opened 2026-09-21
Drop support for Python 3.10 and therefore `cppjit.interactive`

guitargeek:
Python 3.11 is end-of-life in october, and it was the last release that supported the `cppjit.interactive` feature that is now removed.

`cppjit.interactive` enabled the notorious star `*` imports like `from cppjit.interactive import *` to get lazy handles to any C++ scopes. This relied on hacks that were not possible anymore with Python 3.11.


aaronj0 (approved):
Great, thanks!

